PR #34, opened 2026-05-04
chore/frontend setup

larrymijo:
## Summary

Sets up the complete client-side authentication foundation:
- Global auth state via React Context (user + token)
- JWT token persisted in localStorage — survives page refresh
- API service layer isolated from UI components
- ProtectedRoute wrapper redirecting unauthenticated users to /login
- Design tokens (colour palette, typography, spacing) in index.css
- Vite proxy configured to forward /api → http://localhost:3001

---

## Files changed

| File | Purpose |
|---|---|
| `client/src/index.css` | Design tokens — colours, fonts, spacing, radius |
| `client/src/services/authService.js` | Raw fetch calls to POST /api/auth/login and /register |
| `client/src/context/AuthContext.jsx` | Global user + token state, login/register/logout functions |
| `client/src/App.jsx` | AuthProvider wrapper, route definitions, ProtectedRoute |
| `client/vite.config.js` | Proxy /api to localhost:3001 |

---

### ProtectedRoute
Wraps routes that require authentication. Redirects to `/login` with the blocked URL saved in `location.state.from` so the user is sent back after logging in.

| Route | Protected |
|---|---|
| `/profile` | Yes — redirects to /login if no token found |
| All others | No — publicly accessible |

---

## Auth state shape

```js
// What useAuth() returns to any component
{
  user:            { id, email, name, role } | null,
  token:           string | null,
  isAuthenticated: boolean,
  loading:         boolean,
  login:           async (email, password) => user,
  register:        async (email, password, name) => user,
  logout:          () => void,
  getAuthHeader:   () => { Authorization: 'Bearer ...' } | {}
}
```

## How to test

1. `cd client && npm run dev`
2. Open `/profile` — should redirect to `/login`
3. Open DevTools → Application → Local Storage
4. Paste a mock session and refresh — should stay logged in:
```js
localStorage.setItem('hkif_token', 'test')
localStorage.setItem('hkif_user', JSON.stringify({ id:1, email:'test@hkr.se', name:'Test', role:'USER' }))
```
5. Click Log out — localStorage clears, ProtectedRoute redirects again

## Depends on

- `feature/authentication-system` — backend must be running on port 3001 for real API calls to work. Auth state works standalone with mocked localStorage in the meantime.


jasminerezai (approved):
Good stuff overall!

Actually, you've actually knocked out a chunk of Sprint 2 here too (auth context + protected routes). Really appreciate you staying ahead, makes everyone's life easier. :) Approved, green to merge!

client/src/services/AuthService.js
```diff
@@ -0,0 +1,80 @@
+// src/services/authService.js
+//
+// All calls to the auth API live here.
+// AuthContext imports these functions — nothing else should call them directly.
+//
+// Backend base URL is proxied through Vite (see vite.config.js).
+// So /api/auth/login → http://localhost:3001/api/auth/login in dev.
+//
+// Backend response shape (from auth.controller.ts):
```
gunnarwrld, line 9: This matches the current backend implementation in `auth.controller.ts` (we return `{ status: 'success', data: result }` where result is { user, token }). Nicely encoded here, it should plug into `feature/authentication-system` without changes.

client/src/services/AuthService.js
```diff
@@ -0,0 +1,80 @@
...
+    body: JSON.stringify({ email, password }),
+    // JSON.stringify turns { email, password } into a string:
+    // '{"email":"...","password":"..."}'
+  })
+
+  const json = await res.json()
+  // Parse the response body as JSON regardless of status code
+  // so we can read the error message even when res.ok is false
+
+  if (!res.ok) {
```
gunnarwrld, line 42: One idea for later: instead of throwing only `json.message`, we might also include `res.status` or a small error type so the UI can distinguish “wrong credentials” (401) vs “validation error” (400) vs “server error” (500). That will be useful when we add specific messages like “password must be at least 8 characters” from the backend.

client/src/services/AuthService.js
```diff
@@ -0,0 +1,80 @@
...
+// registerRequest
+// Calls POST /api/auth/register
+// Backend expects: { email, password, name }
+// Returns { user, token } on success
+// ─────────────────────────────────────────────────────────────
+export async function registerRequest(email, password, name) {
+  const res = await fetch(`${BASE}/register`, {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify({ email, password, name }),
```
gunnarwrld, line 63: This body matches the backend controller exactly (`registerHandler` expects email, password, name and enforces password length ≥ 8). Good alignment. When we later add more profile fields (e.g. role, maybe student id), we should update both this body and the backend DTO in one go so they stay in sync.

client/src/context/AuthContext.jsx
```diff
@@ -0,0 +1,220 @@
...
+// <AuthProvider>
+//   <Navbar />       ← can call useAuth()
+//   <Routes>...</Routes>  ← pages can call useAuth()
+// </AuthProvider>
+// ─────────────────────────────────────────────────────────────
+export function AuthProvider({ children }) {
+
+  // ── State ──────────────────────────────────────────────────
+  //
+  // user: the logged-in user object or null
```
gunnarwrld, line 47: The documented user shape lines up with what the auth service returns (`{ user, token }`, where user includes id/email/name/role). To keep things robust, it might be nice to link to a backend type/interface (or add a short comment referring to the Prisma model) so we remember to update this context whenever the backend user shape changes.

client/src/context/AuthContext.jsx
```diff
@@ -0,0 +1,220 @@
...
+  // before it finishes reading the stored session.
+  const [loading, setLoading] = useState(true)
+
+  // ── Rehydrate session on page load ─────────────────────────
+  //
+  // useEffect with [] runs once — immediately after the first render.
+  // We check localStorage for a saved token + user.
+  // If found, restore them so the user doesn't have to log in again
+  // every time they refresh the page.
+  useEffect(() => {
```
gunnarwrld, line 71: Really like this TODO. Once we implement `GET /api/users/me` on the backend, we can change this effect to: read token from localStorage → call `/api/users/me` → set `user` from the fresh server data. That will make sure roles/names are always current and not only whatever was saved when the user first logged in.

client/src/context/AuthContext.jsx
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  // ── getAuthHeader ──────────────────────────────────────────
+  // Returns the Authorization header object that any fetch() call
+  // needs when talking to a protected backend endpoint.
+  //
+  // Usage in a service file:
+  //   const { getAuthHeader } = useAuth()
+  //   fetch('/api/schedule', { headers: getAuthHeader() })
+  function getAuthHeader() {
```
gunnarwrld, line 169: Great having this helper!

client/src/App.jsx
```diff
@@ -0,0 +1,115 @@
...
+import { Routes, Route, Navigate, useLocation } from 'react-router-dom'
+import { AuthProvider, useAuth } from './context/AuthContext.jsx'
+
+// ── ProtectedRoute ────────────────────────────────────────────
+// Wraps any route that requires login.
+// If not authenticated → redirect to /login
+// Passes the blocked URL in location.state so LoginPage can
+// redirect back after a successful login.
+function ProtectedRoute({ children }) {
+  const { isAuthenticated, loading } = useAuth()
```
gunnarwrld, line 12: Nice having a `loading` state.  Once we have more loading visuals, we can swap this inline spinner for a shared `<Spinner />`component.

client/vite.config.js
```diff
@@ -0,0 +1,26 @@
...
+    port: 5173,
+    // The port frontend runs on (default)
+
+    proxy: {
+      '/api': {
+        // Any fetch('/api/...') call in your React code gets
+        // forwarded here during development.
+        // So fetch('/api/auth/login') becomes:
+        //   http://localhost:3001/api/auth/login
+        target: 'http://localhost:3001',
```
gunnarwrld, line 18: Confirming that `target: 'http://localhost:3001'` matches our backend dev port. For later deployment, we can introduce `VITE_API_URL` so that the same frontend can talk to the hosted backend without changing this file.

gunnarwrld (approved):
From the backend side, this looks very aligned with feature/authentication-system.

The auth endpoints this client calls (`POST /api/auth/login` and `POST /api/auth/register`) and their payloads match the current auth.controller.ts implementation: the controller requires { email, password } for login and { email, password, name } for register, and both respond with status: 'success' and data: { user, token }. That fits exactly with what AuthService expects and with the user shape documented in AuthContext.

The `getAuthHeader` helper uses the `Authorization: Bearer <token>` format that our auth middleware reads, so future protected endpoints can reuse that.

The loading handling in ProtectedRoute solves the common refresh flicker problem and will make the UX feel solid once we plug in real login/register forms.

A couple of suggestions for future iterations (not blocking):

In AuthService, we could enrich the error handling a bit later so the UI can distinguish between validation errors (e.g. password too short), invalid credentials, and server errors using res.status or a small error type. That will make it easier to show specific messages that match backend validation.

When we add `GET /api/users/me on the backend`, it would be great to switch the localStorage restore flow in `AuthContext` to call that endpoint, so the user object is always derived from the server rather than only from the stored JSON.

Overall, this PR is great and gives us a solid frontend foundation that lines up with the backend auth branch. Once feature/authentication-system is merged into main, this should be able to talk to the real API with minimal changes.

PR #43, opened 2026-05-07
added LoginPage and RegisterPage with validation and JWT auth

### LINKED ISSUE #16: Login & register pages
Build /login and /register pages with forms and validation.
On success: store JWT and redirect to /.
Show error messages on failed attempts.

larrymijo:
## Updated — aligned with PR #44 (auth.ts refactor)

Three changes made in response to review feedback:

**1. getMeRequest implemented (no longer a TODO)**
GET /api/auth/me is now called on page load for session rehydration.
The token is verified against the live DB via authMiddleware before
the session is restored. Invalid/expired tokens are caught and cleared.

**2. Rehydration uses /me — not raw localStorage**
On page refresh: reads token from localStorage → calls /me with
Bearer header → /me verifies token → session restored.
Name/email recovered from saved session after token is verified.

**3. ProfileRole.USER → ProfileRole.MEMBER everywhere**
ROLE_VARIANT in Badge.jsx updated. AuthContext comments updated.
SESSION_KEY replaces USER_KEY for clarity.

Role hierarchy now: MEMBER < LEADER < BOARD_MEMBER < ADMIN

## What this PR does

Implements the /login and /register pages. Both pages use the
shared component library (Card, Input, Button) and the existing
AuthContext — no direct API calls or localStorage access in the
pages themselves.

On successful auth the JWT is stored by AuthContext and the user
is redirected to / (or back to the protected page they came from).
Failed attempts show the server error message in a banner at the
top of the form.

---

## Files added

| File | Purpose |
|---|---|
| `client/src/pages/LoginPage.jsx` | Login form with client-side validation and error display |
| `client/src/pages/RegisterPage.jsx` | Register form with all four fields, password strength indicator |

## Files changed

| File | Change |
|---|---|
| `client/src/App.jsx` | Replaced /login and /register Placeholders with real pages |

---

## Auth flow
User submits form
→ client validation runs (empty fields, email format, password length)
→ if invalid: field errors shown immediately, no API call made
→ if valid: login() or register() called from AuthContext
→ AuthContext calls AuthService.js
→ fetch POST /api/auth/login or /register
→ success: { user, token } returned
→ AuthContext saves to state + localStorage
→ page navigates to / or location.state.from
→ failure: Error thrown with server message
→ page shows error in red banner
---

### Login
| Field | Rule |
|---|---|
| Email | Required, must contain @ and . |
| Password | Required |

### Register
| Field | Rule |
|---|---|
| Name | Required, min 2 characters |
| Email | Required, must contain @ and . |
| Password | Required, min 8 characters (mirrors backend) |
| Confirm | Must match password |

---

## How to test

```bash
git fetch origin
git checkout feat/login-register-pages
cd client && npm run dev
```

1. Go to /login — submit empty form → field errors appear
2. Go to /register — submit empty → all four fields show errors
3. Type 5-char password → strength shows "Too short" in red
4. Type 12-char password → strength shows "Strong" in green
5. Go to /profile without login → redirected to /login
6. Log in → redirected back to /profile

---

## Still to be tested with backend running, I just tested the fronted


JuriRappold:
Also recently realize, we don't want to send the password in plaintext over the internet. And we can't compare two hashes. So we have to encrypt the plaintext password on the frontend, decrypt on the backend, and then we can compare the plaintext password to the stored hash in the DB. 

https://stackoverflow.com/questions/55801772/nodejs-how-to-compare-two-hashes-password-of-bcrypt

larrymijo:
> Also recently realize, we don't want to send the password in plaintext over the internet. And we can't compare two hashes. So we have to encrypt the plaintext password on the frontend, decrypt on the backend, and then we can compare the plaintext password to the stored hash in the DB.
> 
> https://stackoverflow.com/questions/55801772/nodejs-how-to-compare-two-hashes-password-of-bcrypt

Thanks for flagging this — 

I had a look at the backend code and I think we're actually already 
handling this correctly. Here's what's happening:

The password does travel as plaintext in the HTTP request body, but 
it's protected by HTTPS/TLS which encrypts the entire connection — 
that's what the S in HTTPS stands for. So it's not actually going 
over the wire in plaintext in production.

On the backend, auth.service.ts already does:

  const isMatch = await bcrypt.compare(password, user.password)

This is the correct industry standard pattern — the StackOverflow 
link you shared actually explains why this works. bcrypt.compare() 
extracts the salt from the stored hash and uses it to hash the 
incoming plaintext, then compares. That's exactly what's happening.

The approach you're suggesting — encrypting on the frontend and 
decrypting on the backend — would require storing the encryption 
key in our JavaScript, which anyone can read in DevTools. That 
would actually reduce security rather than improve it.

Happy to discuss this further — Does this make sense?

client/src/pages/LoginPage.jsx
```diff
@@ -0,0 +1,288 @@
...
+// useNavigate: returns a function to programmatically change the URL
+// useLocation: reads the current URL — used to send users back to
+//              where they were before being redirected to /login
+
+import { useAuth } from '../context/AuthContext.jsx'
+import { Button, Input, Card } from '../components/ui'
+// Importing from the barrel export (index.jsx) — one clean import line
+
+export default function LoginPage() {
+  const { login } = useAuth()
```
gunnarwrld, line 27: ProfileRole.USER doesn't exists anymore after PR #44. The enum was renamed to MEMBER.
Search the entire `client/` for any hardcoded `'USER'` role string 
and replace with `'MEMBER'`.

client/src/pages/RegisterPage.jsx
```diff
@@ -0,0 +1,297 @@
...
+// Backend expects: POST /api/auth/register { email, password, name }
+// Backend validates: password must be >= 8 characters (auth.controller.ts)
+
+import React, { useState } from 'react'
+import { Link, useNavigate } from 'react-router-dom'
+import { useAuth } from '../context/AuthContext.jsx'
+import { Button, Input, Card } from '../components/ui'
+
+export default function RegisterPage() {
+  const { register } = useAuth()
```
gunnarwrld, line 20: Same as LoginPage: `auth.service.ts` (now `auth.ts`) in PR #44 now defaults new registrations 
to ProfileRole.MEMBER (not USER). Align all frontend role references to the new enum:

  MEMBER < LEADER < BOARD_MEMBER < ADMIN

There is no `USER` role anymore.

client/src/pages/LoginPage.jsx
```diff
@@ -0,0 +1,288 @@
...
+    setLoading(true)
+
+    try {
+      await login(email, password)
+      // login() is from AuthContext. Internally it:
+      //   1. Calls loginRequest(email, password) in AuthService.js
+      //   2. Gets { user, token } back from the backend
+      //   3. Saves user to React state
+      //   4. Saves token to localStorage under 'hkif_token'
+      //   5. Saves user to localStorage under 'hkif_user'
```
gunnarwrld, line 114: The backend `auth.service.ts` login response returns the full Profile row,
which includes profileName. But `req.user` (and thus the JWT payload) now 
only carries `{ id, role }`.

Call `GET /api/auth/me` after login to populate `name/email`, do NOT rely on the login response or localStorage for anything beyond `{ id, role, token }`. Update this comment and verify rehydration logic follows this pattern.

gunnarwrld (changes requested):
Alignment needed with PR #44 (`auth.ts` refactor):

1. `req.user` shape changed. The new authMiddleware only attaches `{ id, role }`.
   The old `protect()` middleware also set name and email on `req.user`.
   `GET /api/auth/me` (getMeHandler) now must be the ONLY source of name/email.
   `AuthService.loginRequest()` and `registerRequest()` should only consume 
   `{ user: { id, role }, token }` from the response, nothing else.

2. The `/me` endpoint is now guarded by authMiddleware (PR #44, `auth.routes.ts`).
   If AuthContext calls `GET /api/auth/me` for session rehydration, it MUST send 
   the Authorization: `Bearer <token>` header. Verify `AuthService.getMeRequest()`
   (or equivalent) does this.

3. `ProfileRole.USER` to `ProfileRole.MEMBER` everywhere in the client.
   Look for 'USER' across client/src/ and replace all role comparisons.

JuriRappold:
> > Also recently realize, we don't want to send the password in plaintext over the internet. And we can't compare two hashes. So we have to encrypt the plaintext password on the frontend, decrypt on the backend, and then we can compare the plaintext password to the stored hash in the DB.
> > https://stackoverflow.com/questions/55801772/nodejs-how-to-compare-two-hashes-password-of-bcrypt
> 
> Thanks for flagging this —
> 
> I had a look at the backend code and I think we're actually already handling this correctly. Here's what's happening:
> 
> The password does travel as plaintext in the HTTP request body, but it's protected by HTTPS/TLS which encrypts the entire connection — that's what the S in HTTPS stands for. So it's not actually going over the wire in plaintext in production.
> 
> On the backend, auth.service.ts already does:
> 
> const isMatch = await bcrypt.compare(password, user.password)
> 
> This is the correct industry standard pattern — the StackOverflow link you shared actually explains why this works. bcrypt.compare() extracts the salt from the stored hash and uses it to hash the incoming plaintext, then compares. That's exactly what's happening.
> 
> The approach you're suggesting — encrypting on the frontend and decrypting on the backend — would require storing the encryption key in our JavaScript, which anyone can read in DevTools. That would actually reduce security rather than improve it.
> 
> Happy to discuss this further — Does this make sense?

Ah ok makes sense.
Also found similiar info here: https://stackoverflow.com/questions/75530827/typescript-asymmetric-encryption.
Though only the public key would be visible in the DevTools and I think the server would read the private key file only on demand ( i think)

Good to have cleared that up! Nice job!

isaqelle:
Looks sooo good! Nice job

client/src/pages/LoginPage.jsx
larrymijo: All three points addressed — pushed in the latest three commits.

1. getMeRequest is now implemented and called on page load for
   rehydration. It sends Authorization: Bearer <token> as required
   by authMiddleware. Invalid tokens are caught and localStorage
   is cleared.

2. After login/register, the full user (including name/email) comes
   from the login/register response (auth.service.ts toUserDto still
   returns them). On rehydration we call /me to verify the token is
   live, then restore name/email from the saved session. The role
   always comes from /me on reload so it reflects any DB changes.

3. ProfileRole.USER replaced with ProfileRole.MEMBER in Badge.jsx
   ROLE_VARIANT and all comments. Searched all of client/src —
   no other hardcoded 'USER' role strings found.

gunnarwrld: Great!

jasminerezai (approved):
Great Lawrence!

PR #44, opened 2026-05-08
refactor(prisma,auth): rename ProfileRole USER to MEMBER, unified auth middleware with role hierarchy

### LINKED ISSUE #41: refactor(schema): replace single ProfileRole enum to support multiple roles per profile
## Problem

`Profile` currently has a single `role: ProfileRole` enum field, which means a person can only hold one role at a time. I was thinking that a leader can simultaneously be a board member (this is not possible with the current design).

## Current schema

```prisma
enum ProfileRole {
  USER
  LEADER
  BOARD_MEMBER
  ADMIN
}

model Profile {
  ...
  role ProfileRole @default(USER)
  ...
}
```

### Option A — Add combined enum value
Add `LEADER_BOARD_MEMBER` as a combined value. No migration complexity but doesn't scale if more role combinations are needed.

### Option B — Replace with join table 
Remove the single `role` field from `Profile` and introduce a `ProfileRoleAssignment` join table:

```prisma
model ProfileRoleAssignment {
  profileId String
  role      ProfileRole

  profile Profile @relation(fields: [profileId], references: [id], onDelete: Cascade)

  @@id([profileId, role])
}
```
Let me know what you guys think or how should we proceed!

gunnarwrld:
### Closes #41

Resolves the schema design discussion from issue #41. Instead of a join table, we adopt a **strict privilege hierarchy** with a single `role` field on `Profile`. Each higher role inherits all permissions of the roles below it:

---

### Usage reference

```ts
// LEADER, BOARD_MEMBER and ADMIN can all create activities
router.post('/activities', authMiddleware, restrictToMinRole('LEADER'), createActivity);

// Only ADMIN
router.delete('/users/:id', authMiddleware, restrictTo('ADMIN'), deleteUser);
```

---

### Notes for reviewers

- `req.user` intentionally only contains `{ id, role }`. If a handler needs `name` or `email`, it must query `prisma.profile.findUnique({ where: { id: req.user.id } })`.
- The old `protect` middleware is fully replaced, do not use both in the same route!!.
- The migration history reset is intentional: the old migrations were generated against a different schema state and would fail on a fresh `prisma migrate deploy`. The new baseline is safe to apply on any clean DB.


server/prisma/schema.prisma
```diff
@@ -21,7 +20,7 @@ enum ActivityStatus {
 }
 
 enum ProfileRole {
```
JuriRappold, line 22:
```suggestion
enum UserRole {
```

server/prisma/schema.prisma
JuriRappold, line 43:
```suggestion
model User {
```

JuriRappold, line 1: Change Table name `Profile` to `User`?
Is more intuitive

JuriRappold:
Noice!! Good Job!

JuriRappold (approved):

server/prisma/schema.prisma
gunnarwrld: I also thought about it, but I'd rather keep `Profile` for a few reasons:

User conflicts with Prisma internals, I think that Prisma reserves `User` as a common model name and some tooling can behave weirdly with it.

Profile is more accurate imo, a User is typically an authenticated identity, while a Profile holds data (name, role, activity history). We may want to split these in the future.

We'd also need to rename `ProfileRole` into `UserRole`, update every `prisma.profile.*` call, all `req.user` type references, and regenerate the Prisma client. 

We could keep discussing it, but I'd rather not take the rename on in this PR.

isaqelle (approved):

PR #48, opened 2026-05-10
fix(server): consolidate Prisma config and repair seed for Prisma

gunnarwrld:
## What & Why

Fixes needed to get the local database initialising cleanly after the Prisma and the `MEMBER` role rename.

---

### 1. Prisma config consolidated (`prisma.config.ts`)

**Problem:** Two config files existed (`prisma.config.ts` and `prisma.config.js`). The `.ts` file was the original Prisma-generated one and used `process.env.DATABASE_URL` directly, which Prisma 7's CLI refuses
to resolve (it requires their own `env()` helper). 

---

### 2. Past migrations removed

**Problem:** Four pre-baseline migration files still existed alongside the
`20260508000000_baseline_with_member_role` migration. On a fresh database,
`prisma migrate deploy` would attempt them first, hit missing tables/columns
from the old schema, and fail.

**Fix:** Deleted the four obsolete migrations. The baseline is the **ONLY**
migration that should ever be applied to a fresh database. Existing
environments that already have these applied are unaffected (the baseline
will simply be skipped by the migration engine).

```
deleted: 20260504181341_prisma_redo
deleted: 20260504181541_prisma_redo_new_tables
deleted: 20260505075728_added_activity_template_notes
deleted: 20260505091925_made_notes_optional
```

---

### 3. Seed fixed (`prisma/seed.ts`)

**Problem 1:** `ProfileRole.USER` was used in the seed but the enum value
was renamed to `MEMBER` in PR #44.

**Problem 2:** When an activity was assigned two time slots, both could
randomly land on the same weekday, violating the
`@@unique([activityId, weekday, startTime])` constraint and crashing the
seed mid-run.

**Fix:** Renamed `USER to MEMBER`. Changed the weekday picker to shuffle all
weekdays first and pick the `i+1` one, guaranteeing distinct weekdays per
activity. Added a `.catch(() => null)` guard as a final safety net.

---

## How to run on a fresh machine

```bash
# 1. Create the database (skip if it already exists)
psql -U <your_pg_user> -c "CREATE DATABASE hkif;" postgres

# 2. Apply the baseline migration
cd server
npx prisma migrate deploy

# 3. Generate the Prisma client
npx prisma generate

# 4. Seed the database
npx prisma db seed
```

> Passwords for seeded users follow the pattern `userXpass`
> (e.g. `user1@example.com` / `user1pass`). User 1 is ADMIN,
> users 2–5 are LEADER, the rest are MEMBER.


jasminerezai (approved):
Looks good to me!

JuriRappold (approved):
Yep Looks good to me!

PR #49, opened 2026-05-10
feat(activities): leader schedule status management + types refactor

### LINKED ISSUE #23: Leader: cancel & delete activity
Allow leaders to delete their own activities.
DELETE /activities/:id on backend.
Show a cancellation notice on the activity card for registered participants. (For the "notice", you can decide if we should have it or not!)

gunnarwrld:
## Summary

Closes #23 by replacing hard-delete behavior with a schedule status update flow, and by cleaning up the shared API typing so controller responses are now typed consistently.

## What changed

Added:

`PATCH /api/activities/:activityId/schedules/:scheduleId/status`

This lets authorized leaders update a session to `CANCELLED`, `DELAYED`, `INACTIVE`, or `ACTIVE` without deleting the schedule row.

Why this approach:
- Preserves participation history.
- Avoids irreversible mistakes.
- Reuses existing `GET /api/schedules` response data so the frontend can show status changes without an extra endpoint.

### Types refactor
`src/types/api.types.ts` was split into domain files:

```text
src/types/
├── express/index.d.ts
├── auth.types.ts
├── activity.types.ts
└── index.ts
```

## Why the types refactor matters

This is not just cleanup.

- Controllers now use `Response<ApiResponse<Dto>>`, so TypeScript can catch response-shape mismatches at compile time.
- `ApiResponse<T>` now matches what controllers actually return, including `status: 'success'`.
- Types are grouped by domain instead of living in one catch-all file.
- Imports are standardized through `../types`, which makes usage simpler and more consistent.
- Adding new DTOs is easier and less conflict-prone.

## Notes

- `LEADER` can only update schedules for activities they own.
- `BOARD_MEMBER` and `ADMIN` can update any activity schedule.
- `src/types/express/index.d.ts` stays separate on purpose because it contains the Express `req.user` augmentation.


JuriRappold:
Just kinda a general question: Which HTTP method are we using for updates? PATCH or PUT? Because, so far I have used PUT, I think thats the methods used in the Project-board. Also don't rn know the difference between the two, only know that PATCH is newer and *might* not be supported by older browsers.

jasminerezai (approved):

jasminerezai:
@JuriRappold the distinction is PUT is used when you update the whole Activity (any field/multiple fields), PATCH is one single field like the status.

server/src/controllers/auth.controller.ts
```diff
@@ -46,7 +53,10 @@ export const loginHandler = asyncHandler(async (req: Request, res: Response) =>
...
 });
 
-export const getMeHandler = asyncHandler(async (req: Request, res: Response) => {
+export const getMeHandler = asyncHandler(async (
+  req: Request,
+  res: Response<ApiResponse<MeResponseDto>>,
+) => {
   // req.user is populated by the protect middleware after token verification
   res.status(200).json({
     status: 'success',
```
JuriRappold, line 62: Do we need the field `status: 'success'`, if we have the HTTP-status code 200?

gunnarwrld: Ye, both serve different purposes. The HTTP status code tells the client the transport succeeded, the request got through and the server responded. 

The `status: 'success'` field in the body is part of our response envelope contract. It's what lets the frontend distinguish a successful data response from an error shape `{ error: string, statusCode: number }` without having to check `res.status` directly every time.

It's also forward-looking, if we ever add partial success states (e.g. a batch operation where some items failed), the body status field gives us room to express that without changing HTTP codes. Stripe and most production APIs do this for exactly that reason.

That said, it's a convention choice.  If you guys prefer to rely purely on HTTP codes and drop `status: 'success'` from the body, that's valid too, but we'd need to do it consistently everywhere. Right now `ApiResponse<T>` in `types/index.ts` has `status: 'success'` established into the interface, so changing it means updating the type and all controllers at once. Happy to do that in a cleanup PR if you guys want!

jasminerezai: Agree with Gunnar here! keeping status: 'success' in the body is the right call. From the frontend side it also makes it much cleaner to handle responses consistently without checking HTTP codes every time. Let's keep it as is, thanks!

PR #50, opened 2026-05-10
Activities CRUD operations

### LINKED ISSUE #10: Activities CRUD endpoints
GET /activities --- public, no auth required
POST /activities --- leader only
PUT /activities/:id --- leader only
DELETE /activities/:id --- leader only

Each activity should include: title, sport, date/time, location, leaderId.

jasminerezai:
Mostly look into:
- activity.controller.ts
- createQueries.ts
- readQueries.ts / getAllActivities
- updateQueries.ts / updateActivity
- activity.routes.ts
- activity.types.ts
- activity.validator.ts

Most of these are files I changed since taking over the ticket.

Note for frontend developers:

Example request for POST `/api/activities`:
```json
{
  "name": "Breathwork",
  "location": "Jasmine's place",
  "description": "(optional, can be left null)",
  "notes": "(optional, can be left null)",
  "maxCapacity": 50, // optional, can be null
  "defaultStatus": "ACTIVE", // optional, can be null
  "leaders": [], // list of profile UUIDs, if empty, the creating user is assigned as leader 
  "timeSlots": [{
    "weekday": "MONDAY",
    "startAt": "18:00:00",
    "endAt": "19:30:00"
  }]
}
```

Example request for PUT `/api/activities/:activityId` (provide any field you want to change):
```json
{
  "maxCapacity": 100,
  "timeSlots": [
    {
      "weekday": "MONDAY",
      "startAt": "16:00:00",
      "endAt": "18:00:00"
    }
  ],
  "leaders": [
    "dd26dbc8-3401-4c38-a31c-9155f21bdc68"
  ]
}
```


gunnarwrld:
This controller needs to adapt to the changes being introduced in PR #49 before this merges.

PR #49 refactors the entire types layer:
- api.types.ts is deleted and replaced with domain-split files
- src/types/index.ts is introduced as the barrel — all consumers import from '../types', not individual files
- asyncHandler is updated to be a generic type adapter (no .catch(next) Express 5 handles that automatically)
- auth.controller.ts is updated to type res as Response<ApiResponse<Dto>> for compile-time response shape checking

Concrete changes needed here:
1. import { Activity } from "../types/activity.types" to import { Activity, TimeSlot } from "../types"
2. Activity and TimeSlot need to be re-exported from src/types/index.ts alongside the types from #49
3. Handler signatures should follow the same pattern as updateScheduleStatusHandler in #49, type res as Response<ApiResponse<YourDto>> so TS catches response shape mismatches at compile time
4. The nested try/catch blocks around Prisma calls should be removed, asyncHandler in #49 is a pass-through (just returns fn), so any throw already reaches errorHandler automatically. Only keep the try/catch around the Zod .parse() call.

server/src/controllers/user.controller.ts
```diff
@@ -0,0 +1,95 @@
+import READ  from '../db/readQueries'
```
JuriRappold: Maybe delete the file? Is probably cleaner this way, if we keep it separate, now that activities & favorites have separate branches.
Cause Im working on the favorite CRUD points rn.

server/src/db/deleteQueries.ts
```diff
@@ -0,0 +1,57 @@
+/*
```
JuriRappold: Same thing with the favorite CRUD points branch stuff.

server/src/db/readQueries.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * @param profileId --> requires to be logged-in, i.e. we need an account
+     * @return Promise<ActivityTemplateModel[]>
+     *     **FAIL**
+     *     --> is empty if the profile doesn't have favorite activities
+     *     **SUCCESS**
+     *     --> array of ActivityTemplateModel objects
+     */
+    static async activitiesFavoritedBy(profileId: string): Promise<ActivityTemplateModel[] | undefined> {
```
JuriRappold: Same branch Favorite CRUD points branch thing

server/src/controllers/activity.controller.ts
```diff
@@ -0,0 +1,134 @@
...
+import { Request, Response } from "express";
+import { asyncHandler } from "../middleware/asyncHandler";
+
+import CREATE from "../db/createQueries";
+import DELETE from "../db/deleteQueries";
+import UPDATE from "../db/updateQueries";
+
+import READ from "../db/readQueries";
+import { CreateActivitySchema, DeleteActivitySchema, UpdateActivityGeneralSchema, UpdateActivityURLSchema } from "../validators/activity.validator";
+import { Activity } from "../types/activity.types";
```
gunnarwrld: Import path will break after PR #49 merges.

`import { Activity } from "../types/activity.types";`

Change to: `import { Activity, TimeSlot } from "../types";`

#49 deletes `api.types.ts` and introduces `src/types/index.ts` as the barrel. The convention going forward is all type imports come from `'../types',` never from the individual files directly. You also need to add `Activity` and `TimeSlot` to the barrel's re-exports in `index.ts`.

server/src/controllers/activity.controller.ts
```diff
@@ -0,0 +1,134 @@
...
+
+
+const newActivity = asyncHandler(
```
gunnarwrld: `asyncHandler` signature changed in PR #49, this usage is now wrong.

#49 rewrites `asyncHandler` as a generic type adapter:
  `export function asyncHandler<P, ResBody, ReqBody, ReqQuery>(fn: ...) { return fn; }`

It no longer wraps with `.catch(next)`. Express 5 forwards async throws automatically, so the wrapper is now purely a TypeScript utility for narrowing `req/res` types.

You could adapt the handlers to follow the pattern from #49's `updateScheduleStatusHandler`:

  `export const newActivity = asyncHandler(async (
    req: Request<{}, ApiResponse<ActivityDto>, Activity>,
    res: Response<ApiResponse<ActivityDto>>,
  ) => { ... });`

JuriRappold (changes requested):
Good Job!! And sorry again for the mess.
If you want to I can pull this branch and remove request changes i made myself, since its my fault :)

server/src/controllers/activity.controller.ts
```diff
@@ -0,0 +1,134 @@
...
+            }
+        }
+
+        try {
+            const data = await CREATE.newActivity(newActivity);
+            res.status(201).json({
+                status: "success",
+                data,
+            });
+        } catch (error) {
```
gunnarwrld: Nested `try/catch` around Prisma calls should be removed.

Because `asyncHandler` in #49 is a pass-through (just returns `fn`), any uncaught throw (including Prisma errors) already goes to `errorHandler` automatically.

Only keep the `try/catch` around the Zod `.parse()` call, that one is intentional because you need to convert a ZodError into an `ApiError.badRequest`.

Same applies to the `try/catch` blocks in `updateActivity`, `deleteActivity`, and `getActivities`.

server/src/types/activity.types.ts
```diff
@@ -0,0 +1,18 @@
+import { Weekday, ActivityStatus } from '../db/prisma';
```
gunnarwrld, line 1: Conflict with PR #49: same filename, different content.

PR #49 already creates src/types/activity.types.ts with:
  - UpdateScheduleStatusBody
  - UpdateScheduleStatusDto

This branch creates src/types/activity.types.ts with:
  - Activity
  - TimeSlot

These will collide on merge.
  - Keep both sets of types in the same file (they're all activity-domain types)
  - Then re-export ALL of them from src/types/index.ts:
      export type { Activity, TimeSlot, UpdateScheduleStatusBody, UpdateScheduleStatusDto } from './activity.types';

Do not leave Activity and TimeSlot out of the barrel or imports across the codebase will be inconsistent.

server/src/routes/activity.routes.ts
```diff
@@ -0,0 +1,32 @@
+import { Router } from 'express';
+import { controller } from "../controllers/activity.controller";
+
+import READ from "../db/readQueries";
+import { protect, restrictTo } from "../middleware/auth.middleware";
```
gunnarwrld: Auth middleware name conflict with PR #49.

PR #49 uses: `import { authMiddleware, restrictToMinRole } from '../middleware/auth'`;
This branch uses: `import { protect, restrictTo } from '../middleware/auth.middleware'`;

Migrate this file to:
  `import { authMiddleware, restrictToMinRole } from '../middleware/auth';`
  `router.post('', authMiddleware, restrictToMinRole('LEADER'), controller.newActivity);`

server/src/routes/activity.routes.ts
```diff
@@ -0,0 +1,32 @@
...
+import { ProfileRole } from "../db/prisma";
+
+export const router: Router = Router(); // express.
```
gunnarwrld: Router export convention conflicts with PR #49.

PR #49 uses: `export default router`
This file uses: `export const router = Router()`;

Maybe we should use default export as it is the standard for route files. (We'll have to update `app.ts` accordingly as ` import activityRoutes from './routes/activity.routes';`

server/src/app.ts
```diff
@@ -1,14 +1,18 @@
...
 import { errorHandler } from './middleware/errorHandler';
+
+// routers
+import {router as scheduleRoutes} from './routes/schedule.routes';
 import authRoutes from './routes/auth.routes';
-=======
-import { errorHandler } from './middleware/errorHandler.js';
->>>>>>> main
+import {router as userRoutes} from './routes/user.routes';
+import {router as activityRoutes} from './routes/activity.routes'
```
gunnarwrld: Route file naming conflict with PR #49.

#49 creates: `src/routes/activities.routes.ts` (plural, default export)
This branch creates: `src/routes/activity.routes.ts` (singular, named export)

Both register under `app.use('/api/activities', ...)`. After merge there will be two route files for the same prefix. We should consolidate into one file  `activities.routes.ts` (plural) to match the REST resource naming and merge the routes from both branches into it.

server/src/db/createQueries.ts
```diff
@@ -0,0 +1,38 @@
...
+                },
+            },
+            include: {
+                leaders: true,
+                timeSlots: true
+            }
+        });
+        return activity;
+    }
+}
```
gunnarwrld, line 38: `addFavorites` is missing but called in `user.controller.ts`

server/src/routes/activity.routes.ts
```diff
@@ -0,0 +1,32 @@
...
+
+
+//get all profiles that have favorited this activity:
+router.get('/favorites', protect, restrictTo(ProfileRole.LEADER, ProfileRole.BOARD_MEMBER, ProfileRole.ADMIN), async (_req, res) => {
+    res.json(await READ.profilesFavorited('6233715d-5631-4e2b-8236-2d39ec323b47'));
```
gunnarwrld: I dont really know if this is on purpose but we should not use Hardcoded UUID in a live route.

`READ.profilesFavorited('6233715d-...')` ignores `req.params` entirely. This also bypasses `asyncHandler`, so any throw here is an unhandled rejection that crashes the server.

Maybe we could turn this into `/:id/favorites`, read `req.params.id`, and route through the controller.

server/src/routes/activity.routes.ts
```diff
@@ -0,0 +1,32 @@
...
+});
+
+// get all participants of an activity, add date parameter
+router.get('/participants', protect, restrictTo(ProfileRole.LEADER, ProfileRole.BOARD_MEMBER, ProfileRole.ADMIN), async (_req, res) => {
```
gunnarwrld: Same issue as above, hardcoded UUID, no `asyncHandler`, crashes on throw.

Fix:`/:id/participants` through the controller.

server/src/controllers/activity.controller.ts
```diff
@@ -0,0 +1,134 @@
...
+        try {
+            newActivity = CreateActivitySchema.parse(req.body);
+        } catch (error) {
+            console.error(error);
+            throw ApiError.badRequest(`Invalid request body: ${error}`);
+        }
+
+        // If no leaders provided, assign the creator as the leader
+        if (!newActivity.leaders || newActivity.leaders.length === 0) {
+            if (req.user?.id) {
```
gunnarwrld: Branch not rebased on `main`, `req.user` has no type here.

`main` already has `server/src/types/express/index.d.ts` which establishes Request with `req.user: { id: string; role: ProfileRole }`. This branch never pulled that in, so every `req.user?.id` access in the controllers has no type backing.

server/src/utils/weekCalculator.ts
```diff
@@ -0,0 +1,161 @@
...
+        if(monthInYear < JANUARY){
+            year--;
+            monthInYear = DECEMBER - Math.abs(monthInYear);
+        }
+        daysInTheMonth = Number(months[monthInYear]);
+        dayInMonth = daysInTheMonth - Math.abs(dayInMonth);
+    }
+
+    return new Date(Date.UTC(year, monthInYear, dayInMonth));
+}
```
gunnarwrld, line 41: I just wanted to say wow, great job @jasminerezai this is amazing!

server/src/controllers/user.controller.ts
gunnarwrld: Agreed, maybe we could delete `user.controller.ts` from this PR entirely. The favorites CRUD belongs on its own branch and will conflict with whatever you're building now anyway. Imo same applies to `user.routes.ts` its better to remove it from this PR and from `app.ts`registration. The `app.use('/api/users', userRoutes)` line in `app.ts` should be commented back out until the favorites branch is ready to merge.

server/src/db/deleteQueries.ts
gunnarwrld: Yep, same here we could remove `deleteFavorites` from `deleteQueries.ts` for this PR and let the `favorites` branch own it.

server/src/db/readQueries.ts
gunnarwrld: Agreed, `activitiesFavoritedBy`and `profilesFavorited` belong on the favorites branch, not here. 

`logIn` should also be removed  cause `auth.service.ts` on main already calls `prisma.profile.findUnique` directly and the refactor to use `READ.logIn` in this branch is unnecessary

 `participantsOf` is activity-domain, so it can stay but fix the typo from `partipantsOf` to `participantsOf`.

gunnarwrld (changes requested):
Hey guyyyyys

Overall the query layer is well structured and it's clear a LOT of work went into this the `READ/CREATE/UPDATE/DELETE` class split is clean and the Zod validation on the activity routes is a solid addition.

That said, this PR needs a few things resolved before it can merge, mostly because of 
conflicts with PR #49 (my branch). Sorry for the mess hehe, I did a types refactor and a 
middleware rename without flagging it early enough, which is now creating friction here. 
I think the refactor will pay off once everything is aligned (typed res bodies catch 
response shape bugs at compile time and the barrel makes imports consistent across the whole codebase), but right now it means you have some rework to do and that's on me, im sorry.

Main blockers:
- `src/types/activity.types.ts` conflicts directly with the same file in #49 (needs to be 
  merged into one coherent file and re-exported through `src/types/index.ts`)
- The auth middleware used here (`protect / restrictTo`) is a different implementation from the one in #49 (`authMiddleware / restrictToMinRole`)
- `asyncHandler` was rewritten in #49 as a generic type adapter, the nested `try/catch` blocks around Prisma calls in the controller can be removed
- `activities.routes.ts` (#49) and `activity.routes.ts` (this branch) both register under 
  /api/activities (these need to be consolidated into one file before both PRs land on main)

Inline comments have the specifics for each of these. Again im so sorry guys and have a goodnight haha

server/src/utils/weekCalculator.ts
jasminerezai: That was Juri hahaha

server/src/routes/activity.routes.ts
jasminerezai: I'm guessing it was Juri's testing code, I'll delete it from here

gunnarwrld (approved):

isaqelle:
Great job on this Jasmine!

gunnarwrld (approved):
Everything looks great! The `weekCalculator` refactor is awesome and letting `setUTCDate()` handle all the overflow logic instead of doing it manually is such a great idea, and it fixes the Sunday bug too. Not sure why the migrations keep sneaking in but other than that everything looks perfect. Amazing job Jasmine!! 🎉

JuriRappold (approved):
Good Job! Hella nice clean-up of the code.

PR #51, opened 2026-05-12
20 favorites    backend endpoints

### LINKED ISSUE #20: Favorites --  backend endpoints
POST /users/me/favorites/:sportId — add favorite sport
DELETE /users/me/favorites/:sportId — remove favorite sport
GET /users/me/favorites — return user's favorite sports list

JuriRappold:
# Implementation of Favorites endpoints
Implementation of Favorites endpoints: the adding, deleting, and receiving of favorites for a logged in User. This also includes input validation for favorites and the minor changes by gunnar.
First the three endpoints: read, create, delete.
Then Example responses on success and failure.
Closes #20 

## Getting a Profiles Favorite
Request: GET request to `/api/users/me/favorites`

Response on success: a 200 w/ an `ApiResponse<ActivityDto[]>` object in the res.body

Response on failure:
	-> a 404 for not found
	-> a 500 for DB issues
	-> a 400 for an invalid profileId input (shouldn't be triggered, the auth middleware should catch before that)

## Sending a new Favorite
Request: POST request to `/api/users/me/favorites/:activityId`

Response on success: a 201 w/ an `ApiResponse<ActivityDto>` object in the res.body

Response on failure:
	-> a 400 for invalid ids, the zod parsing failed (either activity or profile)
	-> a 500 for DB issues

## Deleting a Favorite
Request: DELETE request to `/api/users/me/favorites/:activityId`

Response on success: a 204 response (with no body)

Response on failure:
	-> a 400 for invalid ids, the zod parsing failed (either activity or profile)
	-> a 500 for DB issues

### Sending a new Favorite

```JSON
{
  "status": "success",
  "data": {
    "id": "a6842e57-0be4-4a0f-8e6f-b4ae13520326",
    "name": "Basketball",
    "description": "Basketball session",
    "location": "Location 3",
    "maxCapacity": 14,
    "defaultStatus": "ACTIVE",
    "notes": null,
    "createdAt": "2026-05-11T16:12:24.157Z",
    "updatedAt": "2026-05-11T16:12:24.157Z",
    "leaders": [
      {
        "profileId": "b4624640-26a6-4040-8d3d-5ec0778a0cde",
        "activityId": "a6842e57-0be4-4a0f-8e6f-b4ae13520326"
      }
    ],
    "timeSlots": [
      {
        "id": "e350c4bb-d7b5-49ef-952a-fa3c35649caf",
        "activityId": "a6842e57-0be4-4a0f-8e6f-b4ae13520326",
        "weekday": "MONDAY",
        "startTime": "1970-01-01T14:00:00.000Z",
        "endTime": "1970-01-01T15:00:00.000Z"
      },
      {
        "id": "6e542c12-bdb3-450c-a070-910d08448a6d",
        "activityId": "a6842e57-0be4-4a0f-8e6f-b4ae13520326",
        "weekday": "TUESDAY",
        "startTime": "1970-01-01T15:00:00.000Z",
        "endTime": "1970-01-01T16:00:00.000Z"
      }
    ]
  }
}
```
### Getting your Favorites

```JSON
{
  "status": "success",
  "data": [
    {
      "id": "40cc5b24-4761-42ea-9be5-3a36c262658b",
      "name": "Swimming",
      "description": "Swimming session",
      "location": "Location 4",
      "maxCapacity": 16,
      "defaultStatus": "ACTIVE",
      "notes": null,
      "createdAt": "2026-05-11T16:12:24.158Z",
      "updatedAt": "2026-05-11T16:12:24.158Z",
      "leaders": [
        {
          "profileId": "49f33b55-8bb9-4700-839a-0f9209ab4c4b",
          "activityId": "40cc5b24-4761-42ea-9be5-3a36c262658b"
        }
      ],
      "timeSlots": [
        {
          "id": "4ca4a51b-2f08-4fc9-817c-a0435b9b9549",
          "activityId": "40cc5b24-4761-42ea-9be5-3a36c262658b",
          "weekday": "FRIDAY",
          "startTime": "1970-01-01T17:00:00.000Z",
          "endTime": "1970-01-01T18:00:00.000Z"
        }
      ]
    }
  ]
}
```

### Invalid Id format

```JSON
{
  "error": "Invalid ids: [\n  {\n    \"origin\": \"string\",\n    \"code\": \"invalid_format\",\n    \"format\": \"uuid\",\n    \"pattern\": \"/^([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-8][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}|00000000-0000-0000-0000-000000000000|ffffffff-ffff-ffff-ffff-ffffffffffff)$/\",\n    \"path\": [\n      \"activityId\"\n    ],\n    \"message\": \"Invalid UUID\"\n  }\n]",
  "statusCode": 400
}
```

### No Favorite Activities

```JSON
{
  "error": "You don't have any favorite Activities! :(",
  "statusCode": 404
}
```

### Not an Activity Id

```JSON
{
  "error": "Not found. Invalid id: 00cc5b24-4761-42ea-9be5-3a36c262658b",
  "statusCode": 404
}
```


server/prisma/schema.prisma
```diff
@@ -1,3 +1,4 @@
+// chatgpt refactored schema
```
JuriRappold:
```suggestion
```
Forgot to delete... :)

server/src/db/readQueries.ts
```diff
@@ -1,15 +1,24 @@
...
+export class READ{
+
 
-export class READ {
     /**
      * Purpose: returns the current schedule of the week
      * @return (Schedule&Activity)[] OR undefined ==> see anyWeekSchedule(date: Date) for more details
      */
-    static async currentSchedule(): Promise<ScheduleModel[] | undefined> {
+    static async currentSchedule(): Promise<Schedule[] | undefined>
```
JuriRappold:
```suggestion
```
minor formatting

server/src/db/readQueries.ts
```diff
@@ -1,15 +1,23 @@
...
+export class READ{
+
 
-export class READ {
     /**
      * Purpose: returns the current schedule of the week
      * @return (Schedule&Activity)[] OR undefined ==> see anyWeekSchedule(date: Date) for more details
      */
-    static async currentSchedule(): Promise<ScheduleModel[] | undefined> {
+    {
```
JuriRappold:
```suggestion
    static async currentSchedule(): Promise<ScheduleModel[] | undefined> {
```

server/src/db/readQueries.ts
```diff
@@ -113,10 +166,28 @@ export class READ {
         return activities;
     }
 
-    static async findUserByEmail(email: string) {
```
JuriRappold, line 116: Moved up in my file, so it still exists

server/src/controllers/user.controller.ts
```diff
@@ -0,0 +1,86 @@
...
+        let deletedFavorite: ActivityDto | undefined
+        try {
+            deletedFavorite = await DELETE.deleteFavorite(ids)
+        } catch (error) {
+            throw ApiError.notFound(`Not found. Invalid id: ${ids.activityId}`)
+        }
+        if(!deletedFavorite){
+            throw ApiError.internal(`Something went wrong in the DB: ${deletedFavorite}`);
+        } else{
+            res.status(204).json({status: "success", data: null});
```
jasminerezai: 204 means "No Content", shouldn't send a body with it. Either change to res.status(200).json(...) or res.status(204).send() with no body.

jasminerezai (changes requested):

server/src/controllers/user.controller.ts
```diff
@@ -0,0 +1,86 @@
+import {ApiError} from "../utils/ApiError";
+import {Request, Response} from "express";
+import {asyncHandler} from "../middleware/asyncHandler";
+import {READ, CREATE, DELETE} from '../db/queries';
+import {CreateFavoriteSchema, DeleteFavoriteSchema, isUUID} from "../validators";
+import {ApiResponse, FavoriteCreateDelete, ActivityDto} from "../types";
+
+const getFavorites = asyncHandler(
```
gunnarwrld, line 8: I think the UUID guard on `profileId` is kinda redundant.

`req.user.id` comes from the JWT payload, which was verified already by `authMiddleware` before this handler even runs. If the token is valid, the `id` is a valid UUID by definition (`authMiddleware` would have already thrown `ApiError.unauthorized` if not.)

Myabe is better to remove the entire `if(isUUID(profileId)) / else` block, and leave it like:

  `const profileId = req.user.id;
  const favorites = await READ.activitiesFavoritedBy(profileId);
  if (!favorites.length) throw ApiError.notFound("You don't have any favorite Activities!");
  res.status(200).json({ status: "success", data: favorites });`

The "negative length" else branch is not necesary imo (`Array.length` will never negative).

server/src/controllers/user.controller.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    if(!ids){
+        throw ApiError.internal(`Error was not caught by try-catch`)
+    }
+    else{
+        let newFavorite: ActivityDto
+        try {
+            newFavorite = await CREATE.newFavorite(ids)
+        } catch(error){
+            throw ApiError.notFound(`Not found. Invalid id: ${ids.activityId}`)
```
gunnarwrld: Catching a DB error and re-throwing as `ApiError.notFound` is not the correct approach. A DB error could be a connection issue, a constraint violation, or an internal failure, not necessarily "Not found".

The Prisma error P2025 (record not found on a foreign key) is the only case that maps to 404.
All other Prisma errors should bubble up to `errorHandler` as a 500.
The correct pattern is to NOT wrap Prisma calls in `try/catch` inside the handler.

Let `asyncHandler` forward the rejection to `errorHandler`.

If you need to distinguish P2025:

     `import { Prisma } from '../generated/prisma';
     try {
       newFavorite = await CREATE.newFavorite(ids);
     } catch (error) {
       if (error instanceof Prisma.PrismaClientKnownRequestError && error.code === 'P2025')
         throw ApiError.notFound(`Not found. Invalid id: ${ids.activityId}`);
       throw error; // let errorHandler handle the rest
     }`

server/src/controllers/user.controller.ts
```diff
@@ -0,0 +1,86 @@
...
+    async (req: Request<{ activityId: string; }>, res: Response<ApiResponse<ActivityDto>>) => {
+    let ids: FavoriteCreateDelete | null;
+    try{
+        ids = CreateFavoriteSchema.parse({profileId: req.user.id, activityId: req.params.activityId})
+    } catch(error){
+        console.error(error);
+        throw ApiError.badRequest(`Invalid ids: ${error}`)
+    }
+
+    if(!ids){
```
gunnarwrld: The `if(!ids)` guard after the `try/catch` is unreachable. If Zod parsing throws,the catch block re-throws an `ApiError` before reaching this line. `ids` is never null here. Better to remove it.

server/src/controllers/user.controller.ts
gunnarwrld: Agree here!

server/src/controllers/user.controller.ts
```diff
@@ -0,0 +1,86 @@
...
+        }
+        if(!newFavorite){
+            throw ApiError.internal(`Something went wrong in the DB: ${newFavorite}`);
+        } else{
+            res.status(201).json({status: "success", data: newFavorite});
+        }
+    }
+})
+
+const deleteFavorites = asyncHandler(
```
gunnarwrld, line 28: Same two issues as `newFavorites`, same fixes should also be applied here!

server/src/controllers/user.controller.ts
gunnarwrld: Some HTTP clients silently discard it, but it's technically invalid and can break certain proxies.
As Jasmine said, maybe use `res.status(204).send()`.
The PR description correctly states "a 204 response (with no body)". The implementation contradicts itself haha.

server/src/controllers/user.controller.ts
```diff
@@ -0,0 +1,86 @@
...
+
+});
+
+const newFavorites = asyncHandler(
+    async (req: Request<{ activityId: string; }>, res: Response<ApiResponse<ActivityDto>>) => {
+    let ids: FavoriteCreateDelete | null;
+    try{
+        ids = CreateFavoriteSchema.parse({profileId: req.user.id, activityId: req.params.activityId})
+    } catch(error){
+        console.error(error);
```
gunnarwrld: `console.error(error)` in the Zod catch blocks could lead to a production log leak.

The raw Zod error object will be printed to stdout in production, which can expose internal schema details.

Maybe remove the `console.error` calls. The `ApiError.badRequest` message already surfaces the relevant information to the client.

server/src/db/readQueries.ts
```diff
@@ -71,9 +80,34 @@ export class READ {
...
+            select: {
+                activity: {
+                    include: {
+                        leaders: true,
+                        timeSlots: true
+                    }
+                }
+            }
+        })
+        //unsure about the satisfies keyword here: satisfies ActivityDto[]
```
gunnarwrld, line 104: The map is correct!!

`favorite?.map(a => a.activity)` returns `ActivityDto[]` because the select shape matches.

The optional chaining on `favorite?.` is also unnecessary: `prisma.favorite.findMany()` always returns an array (never `null/undefined`), so `favorite.map(...)` is always safe.

You can replace the optional chain with:
  `return favorites.map(a => a.activity);`
Maybe rename the variable to favorites for clarity?!

server/src/types/favorites.types.ts
```diff
@@ -0,0 +1,3 @@
+import {Favorite} from '../generated/prisma'
+
+export type FavoriteCreateDelete = Pick<Favorite, 'activityId' | 'profileId'>;
```
gunnarwrld: Here i just found a missing newline at end of file (flagged by the github too).
Minor but it'll cause a lint warning in CI and generates a noisy diff on every future edit of this file.
Same issue in `validators/index.ts`.

server/src/validators/favorites.validator.ts
```diff
@@ -0,0 +1,21 @@
+import * as z from 'zod';
+const IdSchema = z.uuid();
+
+
+export const CreateFavoriteSchema = z.object({
```
gunnarwrld: `CreateFavoriteSchema` and `DeleteFavoriteSchema` are identical. Maybe merge them into a single `FavoriteSchema`?! something like:

  `export const FavoriteSchema = z.object({
    profileId: z.uuid(),
    activityId: z.uuid()
  });`

  `export const CreateFavoriteSchema = FavoriteSchema;`
  `export const DeleteFavoriteSchema = FavoriteSchema;`

Or just export `FavoriteSchema` once and use it in both controller handlers.

server/src/db/readQueries.ts
```diff
@@ -94,15 +128,34 @@ export class READ {
...
+
+    /**
+     * returns user based of their unique email
+     * QUESTION: include favorites & participationLog? --> active loading OR lazy loading
+     * @param email string
+     * @return Promise<ProfileModel | undefined>
+     *     ---> ProfileModel: successful query
+     *     ---> undefined: unsuccessful query
+     */
+    static async findUserByEmail(email: string) {
```
gunnarwrld, line 145: Here `findUserByEmail` returns the full Profile row, which includes the password field.
This is flagged as a known risk in the codebase conventions (`schema.prisma`: field is `password` not `passwordHash`).

This function is currently only used in auth (login), where the password is needed for `bcrypt.compare.
That's fine. Since this query isn't used anywhere yet, add a select that excludes password before it gets wired up.

Add a comment:
  // NOTE: returns password hash — only use in auth context, never send response directly

gunnarwrld (changes requested):
Hey Juri! Overall solid PR: the Prisma query design is clean, the barrel/types pattern has been correctly followed, and the route wiring is straightforward. A few things to fix before this merges but that is basically it!

jasminerezai:
Good job Juri and good catches Gunnar!

server/src/routes/user.routes.ts
```diff
@@ -0,0 +1,31 @@
+import { Router } from 'express';
+import { ProfileRole } from '../db/prisma';
+import { controller as userController} from '../controllers/user.controller';
+import { authMiddleware, restrictToMinRole } from "../middleware/auth";
+
+const userRoutes = Router();
+
+//For every route: check if the user is logged In and a member
+userRoutes.use(authMiddleware)
+userRoutes.use(restrictToMinRole(ProfileRole.MEMBER))
```
gunnarwrld: As we talked on discord you can just keep `userRoutes.use(authMiddleware)` and go ahead and delete this line. The individual routes that need stricter roles like LEADER-only endpoints will layer on their own `restrictToMinRole` calls.

server/src/controllers/user.controller.ts
JuriRappold: Yeah, maybe ill add a error type checker. So if its not found it returns 404, else 500

JuriRappold: Removed the try catch and will let it just bubble up to the `errorHandler`

server/src/controllers/user.controller.ts
```diff
@@ -2,53 +2,28 @@ import {ApiError} from "../utils/ApiError";
...
-            res.status(200).json({status: "success", data: favorites})
-        } else { //negative length
-            throw ApiError.internal(`Something went Wrong: favorites.length=${favorites.length}\nfavorites-obj:${favorites}`);
-        }
-    } else{
-        throw ApiError.badRequest(`The provided User has an invalid id: ${profileId}`);
-    }
-
+    const favorites: ActivityDto[] = await READ.activitiesFavoritedBy(profileId);
+    if (!favorites.length) throw ApiError.notFound("You don't have any favorite Activities! :(");
```
jasminerezai: `READ.activitiesFavoritedBy()` already returns an array, and an empty array is still a valid successful response.

With this check, users who simply have no favorites receive a `404` instead of `200` with `data: [ ]`.

I’d recommend removing the throw and letting the normal success response handle empty results as well.

jasminerezai (commented):
Good job Juri, just remove that one line :))

server/src/controllers/user.controller.ts
jasminerezai: I think this could also cause issues for the UI/frontend side.

jasminerezai (approved):

server/src/controllers/user.controller.ts
```diff
@@ -0,0 +1,45 @@
...
+    async (req: Request<{ activityId: string; }>, res: Response<ApiResponse<null>>) => {
+    let ids: FavoriteCreateDelete | null;
+    try{
+        ids = DeleteFavoriteSchema.parse({profileId: req.user.id, activityId: req.params.activityId})
+    } catch(error){
+        throw ApiError.badRequest(`Invalid ids: ${error}`)
+    }
+    const deletedFavorite: ActivityDto = await DELETE.deleteFavorite(ids);
+    if(!deletedFavorite) throw ApiError.internal(`Something went wrong in the DB: ${deletedFavorite}`);
+    res.status(200).json({status: "success", data: null});
```
gunnarwrld: here `deleteFavorites` still responds with `res.status(200)` instead of `res.status(204)`.

gunnarwrld (commented):
Hey Juriii, Big improvement from the last round, the  `getFavorites`  simplification is clean, the duplicate schemas are resolved, and the  `restrictToMinRole`  note was a good catch. There is just one small thing to fix and that's it!

server/src/controllers/user.controller.ts
JuriRappold: yeah, but I have a response body. I initially had `res.status(204).json({status: "success", data: null});`, but then Jasmine correctly pointed out that the 204 response code doesn't have a body. I used the same format as Jasmine did for the activity deletion endpoint. So either we have: `res.status(200).json({status: "success", data: null})`; or `res.status(204).send();`, we should stick to either or.

JuriRappold: Note the new commit

jasminerezai (approved):
Looks good overall! Just waiting on the 200 or 204 decision above and then we're good to merge

server/src/controllers/user.controller.ts
jasminerezai: I think we should keep it `res.status(200).json({status: "success", data: null})` because when you make a request it's nicer to see a response instead of nothing. @gunnarwrld

JuriRappold: But a `res.status(204).send()` does the same thing. I think I prefer the 204-status-code. They're both equivalent in meaning, but I think it might be easier to just parse the status code out of the response in the frontend to confirm deletion.

And sending the activity, I'd argue, doesn't make a lot of sense, since the user would have to click on the activity in some way, shape, or form, meaning the user already has the activity --> no need to send it again. (so a `res.status(200).json({status: "success", data: deleted})` doesn't make any sense tbh. @gunnarwrld

server/src/controllers/user.controller.ts
```diff
@@ -26,7 +26,7 @@ const newFavorites = asyncHandler(
 })
 
 const deleteFavorites = asyncHandler(
-    async (req: Request<{ activityId: string; }>, res: Response<ApiResponse<null>>) => {
+    async (req: Request<{ activityId: string; }>, res: Response<ApiResponse<ActivityDto>>) => {
```
JuriRappold, line 29:
```suggestion
async(req: Request<{ activityId: string; }>, res: Response) => {
```

server/src/controllers/user.controller.ts
```diff
@@ -35,7 +35,7 @@ const deleteFavorites = asyncHandler(
     }
     const deletedFavorite: ActivityDto = await DELETE.deleteFavorite(ids);
     if(!deletedFavorite) throw ApiError.internal(`Something went wrong in the DB: ${deletedFavorite}`);
-    res.status(200).json({status: "success", data: null});
+    res.status(200).json({status: "success", data: deletedFavorite});
```
JuriRappold:
```suggestion
res.status(204).send();
```

JuriRappold (commented):
@gunnarwrld just need your feedback :)

server/src/controllers/user.controller.ts
jasminerezai: Yeah, that’s fair. I mainly just wanted to point out that a `204` response can’t include a body. So either:

```
res.status(204).send()
```

or

```
res.status(200).json({ status: "success", data: null })
```

are both valid, we should just stay consistent across the API.

jasminerezai:
Got OK from Juri. Made the 204 fix on your branch since he's busy atm. Merging now so we're ready for today's presentation

PR #52, opened 2026-05-12
feat: add leader create & edit activity form

### LINKED ISSUE #22: Leader: create & edit activity form
Build a form (visible to leaders only) with fields:
title, sport, date, time, location, max capacity.

POST to /activities on create, PUT to /activities/:id on edit.
Redirect to activity detail page on success.

jasminerezai:
Closes #22

Added ActivityFormPage.jsx with:
- Create and edit modes
- Role check (LEADER, BOARD_MEMBER, ADMIN only)
- All fields: name, location, description, notes, maxCapacity, defaultStatus, timeSlots
- POST /api/activities on create
- PUT /api/activities/:id on edit
- Redirects to activity detail page on success

### DIFF
--- a/client/src/App.jsx
+++ b/client/src/App.jsx
@@ -93,8 +93,8 @@ function AppRoutes() {
           <Route path="/login"          element={<LoginPage />} />
           <Route path="/register"       element={<RegisterPage />} />
           <Route path="/activities" element={<ActivitiesPage />} />
-          <Route path="/activities/:id" element={<Placeholder title="Activity Detail" />} />
           <Route path="/activities/new" element={<ProtectedRoute><ActivityFormPage /></ProtectedRoute>} />
+          <Route path="/activities/:id" element={<Placeholder title="Activity Detail" />} />
           <Route path="/activities/:id/edit" element={<ProtectedRoute><ActivityFormPage /></ProtectedRoute>} />
 
           {/* Protected route — must be logged in */}
--- a/client/src/App.jsx
+++ b/client/src/App.jsx
@@ -6,6 +6,7 @@ import Navbar from './components/Navbar.jsx'
 import ActivitiesPage from './pages/ActivitiesPage.jsx'
 import LoginPage    from './pages/LoginPage.jsx'
 import RegisterPage from './pages/RegisterPage.jsx'
+import ActivityFormPage from './pages/ActivityFormPage.jsx'
 
 // ── ProtectedRoute ────────────────────────────────────────────
 // Wraps any route that requires login.
@@ -93,6 +94,8 @@ function AppRoutes() {
           <Route path="/register"       element={<RegisterPage />} />
           <Route path="/activities" element={<ActivitiesPage />} />
           <Route path="/activities/:id" element={<Placeholder title="Activity Detail" />} />
+          <Route path="/activities/new" element={<ProtectedRoute><ActivityFormPage /></ProtectedRoute>} />
+          <Route path="/activities/:id/edit" element={<ProtectedRoute><ActivityFormPage /></ProtectedRoute>} />
 
           {/* Protected route — must be logged in */}
           <Route
--- a/client/src/pages/ActivityFormPage.jsx
+++ b/client/src/pages/ActivityFormPage.jsx
@@ -12,18 +12,6 @@ export default function ActivityFormPage() {
   const { id: activityId } = useParams()
   const isEditMode = Boolean(activityId)
 
-  // Role check
-  const canManageActivities = ['LEADER', 'BOARD_MEMBER', 'ADMIN'].includes(user?.role)
-  if (!canManageActivities) {
-    return (
-      <div style={{ padding: '48px 24px', textAlign: 'center' }}>
-        <p style={{ color: 'var(--color-danger)' }}>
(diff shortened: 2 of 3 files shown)


client/src/App.jsx
```diff
@@ -93,6 +94,8 @@ function AppRoutes() {
           <Route path="/register"       element={<RegisterPage />} />
           <Route path="/activities" element={<ActivitiesPage />} />
           <Route path="/activities/:id" element={<Placeholder title="Activity Detail" />} />
+          <Route path="/activities/new" element={<ProtectedRoute><ActivityFormPage /></ProtectedRoute>} />
```
gunnarwrld, line 96: Route order bug, `/activities/new` will never match.

React Router v6 matches routes in declaration order. 

When the user navigates to `/activities/new`, React Router matches
`/activities/:id` first and passes `id="new"` to the Placeholder component
before it ever reaches the `/activities/new` route.

Declare `/activities/new` BEFORE `/activities/:id`:

client/src/pages/ActivityFormPage.jsx
```diff
@@ -0,0 +1,329 @@
...
+const STATUSES = ['ACTIVE', 'INACTIVE', 'CANCELLED', 'DELAYED']
+
+export default function ActivityFormPage() {
+  const { user, getAuthHeader } = useAuth()
+  const navigate = useNavigate()
+  const { id: activityId } = useParams()
+  const isEditMode = Boolean(activityId)
+
+  // Role check
+  const canManageActivities = ['LEADER', 'BOARD_MEMBER', 'ADMIN'].includes(user?.role)
```
gunnarwrld: Role check runs as an early return inside a component that already has `useState/useEffect` hooks declared below it.

Hooks must be called unconditionally on every render, in the same order.

When `canManageActivities` is false the component `returns` before reaching `useState/useEffect`,
which means React sees a different number of hook calls between renders and will throw:

Move all `useState/useEffect` declarations ABOVE the role check, then do the early return:

client/src/pages/ActivityFormPage.jsx
```diff
@@ -0,0 +1,329 @@
...
+    { weekday: 'MONDAY', startAt: '18:00:00', endAt: '19:30:00' }
+  ])
+
+  // UI state
+  const [loading, setLoading] = useState(false)
+  const [errors, setErrors] = useState({})
+  const [serverError, setServerError] = useState('')
+
+  // Load existing activity in edit mode
+  useEffect(() => {
```
gunnarwrld, line 32: The fetch in the edit-mode useEffect has no auth header.

`fetch(/api/activities/${activityId})`

If `GET /api/activities/:id` requires authentication (`authMiddleware` on the activities router),
this will return 401 silently, the `.then(json => ...)` won't throw, it'll just receive
`{ error: "No token", statusCode: 401 }` and the form will stay empty with no feedback to the user.

Maybe doing this makes more sense?!(correct me if i'm wrong):
  `fetch(/api/activities/${activityId}, {
    headers: { ...getAuthHeader() }
  })`

client/src/pages/ActivityFormPage.jsx
```diff
@@ -0,0 +1,329 @@
...
+
+      const json = await res.json()
+
+      if (!res.ok) {
+        setServerError(json.error || 'Something went wrong.')
+        return
+      }
+
+      // Redirect to activity detail page on success
+      navigate(`/activities/${json.data.id}`)
```
gunnarwrld: `navigate(/activities/${json.data.id})` will navigate to the Placeholder page, not a real Activity Detail page. That's fine for now, but there's a second issue: in edit mode the response from `PUT /api/activities/:id` might not include `json.data.id` depending on the backend response shape.

Check what PUT returns, if it returns `{ status: 'success', data: <updated activity> }`
this is fine. If the shape differs `json.data.id` will be undefined and the redirect will go to `/activities/undefined`.

client/src/pages/ActivityFormPage.jsx
```diff
@@ -0,0 +1,329 @@
...
+                value={defaultStatus}
+                onChange={e => setDefaultStatus(e.target.value)}
+                disabled={loading}
+                style={{
+                  padding: '9px 13px',
+                  border: '1.5px solid var(--color-border)',
+                  borderRadius: 'var(--radius-sm)',
+                  fontSize: 'var(--text-base)',
+                  fontFamily: 'var(--font-body)',
+                  background: '#ffffff',
```
gunnarwrld: `background: '#ffffff'` is hardcoded on both `<select>` elements.
This will break the dark mode (if we ever implement one lmao) the select will have a white background regardless of theme.

Maybe its better to use a: `background: 'var(--color-surface)'` instead.

gunnarwrld (changes requested):
Hey Jasmineee, the overall shape of this pr is good ,the `create/edit` mode split, the form hydration, the validation flow, and the loading state handling are all done correctly. Just some things to fix before merge though.

larrymijo (approved):

isaqelle (approved):

gunnarwrld (approved):
Great Jasmine it looks awesomeeee, ready to merge!!🎉

PR #58, opened 2026-05-13
Updated schedule-GET routes to the current standard

### LINKED ISSUE #55: Schedules Endpoint Update --- Type Checks
# What already exists:
- GET `/api/schedules/current` --> returns the current weeks schedule

- GET `/api/schedules?date=<some_date(default: today)&entireWeek=<boolean(default: true)>` --> returns activities of any given date and the entire week (if you want to)
	- if you just send a GET `/api/schedules` with no parameters the current weeks schedule is also returned

# Updating
- new Request parameter in the `asyncHandler`
- new type of Response body

---
# Creating a new weeks schedule
I'm unsure on how we want to do this, preferably we automate this. But I'm unsure on how to do this, we could use cron-scheduling, but that is another abstraction layer. Otherwise, we could once a day check if its Monday and then create a new schedule for the week two weeks from that Monday (as a bit of a buffer).

JuriRappold:
Updated both get requests to the current standard:

GET `/api/schedules/current` --> returns the currents weeks schedule, is a wrapper around the second GET-request

GET `/api/schedules?date=<some-date>&entireWeek=<boolean>`

**Valid `date` values**
Dates of the format *YYYY-MM-dd* or the string "today"
[See also zod.iso.date() type](https://zod.dev/api#iso-dates)

> Defaults to the current date

**Valid `entireWeek` values**
Any typical boolean values, I would prefer if we keep it to just *true* and *false*
[See also zod.stringbool()](https://zod.dev/api#stringbool)
> Defaults to *true*

> In General the route defaults to the currents weeks schedule

### DIFF
--- a/server/src/controllers/schedule.controller.ts
+++ b/server/src/controllers/schedule.controller.ts
@@ -3,7 +3,7 @@ import {ApiError} from "../utils/ApiError";
 import {Request, Response} from "express";
 import {asyncHandler} from "../middleware/asyncHandler";
 import {ApiResponse} from "../types";
-import { ScheduleDto } from '../types/schedule.types';
+import {ScheduleDto} from '../types'
 import {ScheduleDateSchema, ScheduleBoolWeekSchema} from "../validators/schedule.validator";
 import {output, ZodISODate, ZodLiteral, ZodNullable, ZodOptional, ZodUnion} from "zod";
 
--- a/server/src/controllers/schedule.controller.ts
+++ b/server/src/controllers/schedule.controller.ts
@@ -4,6 +4,8 @@ import {Request, Response} from "express";
 import {asyncHandler} from "../middleware/asyncHandler";
 import {ApiResponse} from "../types";
 import { ScheduleDto } from '../types/schedule.types';
+import {ScheduleDateSchema, ScheduleBoolWeekSchema} from "../validators/schedule.validator";
+import {output, ZodISODate, ZodLiteral, ZodNullable, ZodOptional, ZodUnion} from "zod";
 
 const currentWeek = asyncHandler(
     async (_req: Request, res: Response< ApiResponse< ScheduleDto[] > >) => {
@@ -18,43 +20,35 @@ const currentWeek = asyncHandler(
  */
 const getSchedule = asyncHandler(
     async (req: Request, res: Response< ApiResponse< ScheduleDto[] > >) => {
-        if(!req.query){
-            const thisWeek: ScheduleDto[] = await READ.currentSchedule();
-            res.status(200).json({status: "success", data: thisWeek});
+        let schedule: ScheduleDto[];
+        let date: output<ZodOptional<ZodNullable<ZodUnion<readonly [ZodISODate, ZodLiteral<"today">]>>>> | Date;
+        let entireWeek: boolean | null;
+        const resultDate = ScheduleDateSchema.safeParse(req.query.date);
+        const resultBool = ScheduleBoolWeekSchema.safeParse(req.query.entireWeek);
+        if(!resultDate.success) throw ApiError.badRequest(`Invalid Query Values: Bad date ${req.query.date}`);
+        else { date = resultDate.data}
+
+        if(!resultBool.success) throw ApiError.badRequest(`Invalid Query Values: Bad Boolean ${req.query.entireWeek}`);
+        else{ entireWeek = resultBool.data}
+
+
+        if (!date || date === "today") {
+            date = new Date();
+            date = date as Date;
+        } else{
+            date = new Date(date);
+        }
+        date.setUTCHours(0, 0, 0, 0);
+        if( typeof entireWeek === "undefined"){ entireWeek = true}
+        if (entireWeek) {
+            schedule = await READ.anyWeekSchedule(date);
+            res.status(200).json({status: "success", data: schedule});
             return;
         }
         else {
-            let schedule: ScheduleDto[];
-            let date: string | Date | null = req.query.date as (Date | string | null);
-            let entireWeek: string | boolean = req.query.entireWeek as string;
-
-            if (!date || date === "today") {
-                date = new Date();
-            }
-            else {
-                if (typeof date === "string"){ // or better check for "YYYY-MM-dd" format?
-                    date = new Date(date)
-                }
-            }
-            date.setUTCHours(0, 0, 0);
-
-            if (typeof entireWeek === "undefined") {
-                entireWeek = true;
-            }
-            if ( typeof entireWeek === "string"){
-                entireWeek = entireWeek === "true";
-            }
-
-            if (entireWeek) {
-                schedule = await READ.anyWeekSchedule(date);
-                res.status(200).json({status: "success", data: schedule});
-                return;
-            }
-            else {
-                schedule = await READ.anyDaySchedule(date) ?? [];
-                res.status(200).json({status: "success", data: schedule});
-                return;
-            }
+            schedule = await READ.anyDaySchedule(date) ?? [];
+            res.status(200).json({status: "success", data: schedule});
+            return;
         }
     }
 )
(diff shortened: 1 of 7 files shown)


server/src/controllers/schedule.controller.ts
```diff
@@ -22,58 +17,43 @@ const currentWeek = asyncHandler(
  * route: '/api/schedules?date=<someDate>&entireWeek=<boolean>'
  */
 const getSchedule = asyncHandler(
-    async (req: Request, res: Response) => {
+    async (req: Request, res: Response< ApiResponse< ScheduleDto[] > >) => {
         if(!req.query){
```
jasminerezai: This condition is always false. Express never sets req.query to null or undefined, it's always at least an empty object {}. This block of code will never run. Delete it.

server/src/controllers/schedule.controller.ts
```diff
@@ -22,58 +17,43 @@ const currentWeek = asyncHandler(
...
             if (!date || date === "today") {
-                date = new Date(new Date().getUTCFullYear(), new Date().getUTCMonth(), new Date().getUTCDate());
+                date = new Date();
             }
             else {
                 if (typeof date === "string"){ // or better check for "YYYY-MM-dd" format?
                     date = new Date(date)
                 }
             }
+            date.setUTCHours(0, 0, 0);
```
jasminerezai: Minor issue: setUTCHours(0,0,0) leaves existing milliseconds intact. Safer to use setUTCHours(0,0,0,0) to fully normalize the timestamp.

server/src/controllers/schedule.controller.ts
```diff
@@ -22,58 +17,43 @@ const currentWeek = asyncHandler(
...
+            let date: string | Date | null = req.query.date as (Date | string | null);
             let entireWeek: string | boolean = req.query.entireWeek as string;
 
             if (!date || date === "today") {
-                date = new Date(new Date().getUTCFullYear(), new Date().getUTCMonth(), new Date().getUTCDate());
+                date = new Date();
             }
             else {
                 if (typeof date === "string"){ // or better check for "YYYY-MM-dd" format?
                     date = new Date(date)
```
jasminerezai: If someone passes a bad date like `?date=banana`, `new Date("banana")` won't throw, it silently creates an `Invalid Date` object instead. That can later cause confusing failures in the DB query layer and return a generic 500.

Consider validating immediately after parsing:

```ts
if (isNaN(date.getTime())) {
    throw ApiError.badRequest("Invalid date.");
}
```

server/src/db/readQueries.ts
```diff
@@ -16,41 +17,34 @@ export class READ {
...
-                },
-                orderBy: {
-                    startAt: "asc"
-                },
-                include: {
-                    activity: true
-                }
-            });
-        }
-        if (schedule?.length === 0) {
```
jasminerezai, line 55: Before this PR, an empty week returned a 404. Now it returns a 200 with an empty array. Was this intentional or a side effect of the refactor? Both are fine, just want to confirm it's deliberate since it changes the API behavior.

server/src/controllers/schedule.controller.ts
```diff
@@ -22,58 +17,43 @@ const currentWeek = asyncHandler(
...
             if (typeof entireWeek === "undefined") {
                 entireWeek = true;
             }
             if ( typeof entireWeek === "string"){
-                if( entireWeek ===  "true"){
-                    entireWeek = true;
-                } else{
-                    entireWeek = false
-                }
+                entireWeek = entireWeek === "true";
```
jasminerezai: Right now any value other than `"true"` becomes `false`, so values like `?entireWeek=banana` are silently accepted.

Consider validating explicitly and returning a `400` for invalid boolean values instead.

jasminerezai (changes requested):
Great job Juri! Left some comments

server/src/types/schedule.types.ts
```diff
@@ -0,0 +1,6 @@
+import { Schedule, ActivityTemplate } from '../generated/prisma';
+// import {Weekday, ActivityStatus} from '../db/prisma'
+
+export type ScheduleDto = Schedule &{
```
gunnarwrld, line 4: Add this type to the types barrel (`src/types/index.ts`).

So we could be able to import it like this:
 `import { ScheduleDto } from '../types'`

Add to `index.ts`:
  `export type { ScheduleDto } from './schedule.types';`

And update both import sites to use `'../types'` instead of `'../types/schedule.types'`.

server/src/controllers/schedule.controller.ts
gunnarwrld: Agree here! This branch is unreachable.

So the entire `if(!req.query)` block and its `READ.currentSchedule()` call inside `getSchedule`
will never execute. The route always falls into the else branch.

This is a logic bug: if the user hits `GET /api/schedules` with no query params,
they expect the current week's schedule but instead they'll hit the else branch,
find date and `entireWeek` both undefined, and still get the current week by coincidence
(because the defaults kick in). It works by accident, not by design lmao.

Let the defaults handle the empty case.

server/src/controllers/schedule.controller.ts
```diff
@@ -2,18 +2,13 @@ import {READ}  from '../db/readQueries'
...
-        const schedule = await READ.currentSchedule();
-        if(schedule){
-            res.status(200).json(schedule)
-        }
-        else{
-            // res.send(ApiError.notFound("The Week you are requesting hasn't been scheduled yet!"))
-            throw ApiError.notFound("The Week you are requesting hasn't been scheduled yet!")
-        }
+    async (_req: Request, res: Response< ApiResponse< ScheduleDto[] > >) => {
+        const schedule: ScheduleDto[] = await READ.currentSchedule();
```
gunnarwrld, line 12: `READ.currentSchedule()` now returns `ScheduleDto[]` (never undefined), so the
old `notFound` throw is perfectly removed. But there's no empty-array handling if the current week has no scheduled activities, the response will be:
  `{ status: "success", data: [] }`

That's valid REST behaviour (200 with empty array is correct), just flagging it
as a deliberate decision so the frontend team knows to handle the empty array case. (@isaqelle @larrymijo)

server/src/controllers/schedule.controller.ts
```diff
@@ -2,18 +2,15 @@ import {READ}  from '../db/readQueries'
...
```
jasminerezai, line 12: Hey! Just a heads up - the response shape here changed to { status: 'success', data: [...] } which means Isabelle's frontend fetch will break since she's checking Array.isArray(data) which will now always be false. Your PR should merge first, then Isabelle fixes her fetch. I'll leave a comment on her PR too.

jasminerezai (approved):
Looks good! Left one comment about coordination with Isabelle's PR on line R12, but otherwise this is clean. Nice refactor Juri

jasminerezai (commented):
Looks good! Left one comment about coordination with Isabelle's PR on line R12, but otherwise this is clean. Nice refactor Juri

server/src/db/readQueries.ts
JuriRappold: Its deliberate change, since the query can return an empty result, which is a valid result.
The query returned successfully with the result. And since it is a success, 200 i think is a valid.

JuriRappold:
Note that i added the missing leaders. I noticed this when I looked at Isabelles PR that I forgot to include the activity leaders.

server/src/types/schedule.types.ts
jasminerezai: Gunnar is right, can you add this before merging Juri? It's a small fix but keeps our imports consistent

jasminerezai (changes requested):
Good job again Juri! 

Add export type { ScheduleDto } from './schedule.types' to src/types/index.ts and update the two import sites to use '../types' instead of '../types/schedule.types'

jasminerezai:
Resolved merge conflicts and fixed the ScheduleDto barrel export. Merging now

PR #60, opened 2026-05-15
#21 Favorite Activities UI

### LINKED ISSUE #21: Favorites -- UI on profile page
Add a favorite/heart button to each activity card on the Activities page.

- Logged-in users can click the heart to save/unsave an activity as a favorite
- The heart should toggle visually (filled/unfilled) based on current state
- Favorites should be stored (backend endpoint TBD)
- Non-logged-in users should not see the heart button OR maybe they can see it but it should say that they need to be logged in

isaqelle:
## Summary

Added favorites functionality to the Activities page with backend integration and authenticated favorite toggling.

## Features

* Added heart/favorite button to each activity card
* Users can favorite/unfavorite activities
* Favorites are fetched from backend on page load
* Connected frontend to favorites API endpoints
* Added JWT-authenticated requests for favorites actions
* Redirect unauthenticated users to login when attempting to favorite

## Notes

* Favorite state updates immediately in the UI

### DIFF
--- a/client/src/pages/ActivitiesPage.jsx
+++ b/client/src/pages/ActivitiesPage.jsx
@@ -2,6 +2,8 @@ import React, { useState, useEffect } from 'react'
 import { useNavigate } from 'react-router-dom'
 import { useAuth } from '../context/AuthContext.jsx'
 import { fetchFavorites, addFavorite, removeFavorite, } from '../services/FavoritesService.js'
+import Card from '../components/ui/Card.jsx'
+import Button from '../components/ui/Button.jsx'
 
 export default function ActivitiesPage() {
   // ── API State ─────────────────────────────────────────────
@@ -141,15 +143,12 @@ export default function ActivitiesPage() {
           <p>No activities found.</p>
         ) : (
           filteredActivities.map(activity => (
-            <div
+            <Card
               key={activity.id}
+              padding="md"
+              shadow="sm"
               style={{
                 position: 'relative',
-                border: '1px solid var(--color-border)',
-                borderRadius: '12px',
-                padding: '20px',
-                background: 'var(--color-surface-raised)',
-                boxShadow: '0 2px 6px rgba(0,0,0,0.05)',
               }}
             >
               {/* Activity title */}
@@ -215,7 +214,7 @@ export default function ActivitiesPage() {
                   ))}
                 </div>
               )}
-            </div>
+            </Card>
           ))
         )}
       </div>
--- a/client/src/pages/ActivitiesPage.jsx
+++ b/client/src/pages/ActivitiesPage.jsx
@@ -1,17 +1,38 @@
 import React, { useState, useEffect } from 'react'
+import { useNavigate } from 'react-router-dom'
+import { useAuth } from '../context/AuthContext.jsx'
 
 export default function ActivitiesPage() {
   // ── API State ─────────────────────────────────────────────
   const [activities, setActivities] = useState([])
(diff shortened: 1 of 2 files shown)


client/src/pages/ActivitiesPage.jsx
```diff
@@ -84,6 +133,84 @@ export default function ActivitiesPage() {
               }}
             >
               {/* Activity title */}
+              {/* Heart-button */}
+              <button
+                onClick={async () => {
```
larrymijo: Extract the button logic — the whole onClick async handler inside the JSX is getting long. Worth pulling it out into a named function like handleToggleFavorite(activityId) above the return. Makes it a lot easier to read and test.

client/src/pages/ActivitiesPage.jsx
```diff
@@ -84,6 +133,84 @@ export default function ActivitiesPage() {
...
+                    navigate('/login')
+                    return
+                  }
+
+                  try {
+
+                    // Remove favorite
+                    if (favoriteActivities.includes(activity.id)) {
+
+                      const response = await fetch(
```
larrymijo: Move fetch calls to a service — the add/remove fetch calls are inline in the JSX. Since we already have AuthService.js, it'd be cleaner to either add them there or create a FavoritesService.js. Keeps the component focused on rendering.

client/src/pages/ActivitiesPage.jsx
```diff
@@ -76,6 +124,7 @@ export default function ActivitiesPage() {
             <div
```
larrymijo: worth using existing Card component

larrymijo (commented):
Nice work overall, the optimistic toggle and auth flow are solid! Just left some comments as suggestions for the button handler, move calls to a service and reusing our existing Card/Button components, but nothing blocking.

client/src/pages/ActivitiesPage.jsx
isaqelle: Yup agreed, it's pretty long. I'll make a function for it

client/src/pages/ActivitiesPage.jsx
```diff
@@ -84,6 +133,84 @@ export default function ActivitiesPage() {
...
+                            Authorization: `Bearer ${token}`,
+                          },
+                        }
+                      )
+
+                      if (!response.ok) {
+                        throw new Error('Failed to remove favorite')
+                      }
+
+                      setFavoriteActivities(
```
jasminerezai: Optimistic update runs before the request here. If the DELETE fails, the heart disappears from the UI but the favorite still exists in the backend. Save the previous state before this line and restore it in the catch block if the request fails.

client/src/pages/ActivitiesPage.jsx
```diff
@@ -31,6 +37,48 @@ export default function ActivitiesPage() {
...
+
+      } catch (error) {
+
+        console.error('Failed to fetch favorites:', error)
+      }
+    }
+
+    fetchFavorites()
+
+}, [isAuthenticated])
```
jasminerezai: Missing token in the dependency array. If the token changes without isAuthenticated changing, favorites won't re-fetch. Change to [isAuthenticated, token]

jasminerezai (changes requested):
Nice work! :)) left comments on lines 80 and 164. Everything else looks good and I agree with Lawrence's comments too

larrymijo (approved):

jasminerezai (approved):
Good job with the changes!

PR #63, opened 2026-05-15
feat: add participation register and unregister endpoints

### LINKED ISSUE #18: Participation endpoints -- register & unregister
POST /activities/:id/participate — register logged-in user
DELETE /activities/:id/participate — unregister

Check capacity before registering. Return 409 if full.
Return updated participant count.

I’m not sure how we want to handle capacity. Do we have a maximum capacity for activities at  uni? This is just a suggestion. Let me know what you think.

jasminerezai:
Closes #18 
 
 Added register and unregister endpoints for activity participation.
  
  - POST /api/activities/:activityId/schedules/:scheduleId/participate
  - DELETE /api/activities/:activityId/schedules/:scheduleId/participate
  
Checks capacity before registering. Returns 409 if full or already registered. Returns updated participant count.


server/src/controllers/activities.controller.ts
gunnarwrld, line 1: The capacity check and the participation insert are two separate queries with no transaction.

  ```const count = await READ.participantCount(scheduleId)      // query 1
  if (count >= schedule.activity.maxCapacity) throw ...
  // ← gap here, another request can slip through
  await CREATE.registerParticipation(profileId, scheduleId)  // query 2
```
Two concurrent requests can both pass the capacity check before either inserts,
resulting in over-capacity registrations.

Maybe wrap in a transaction and use a count check inside it like:

```
  await prisma.$transaction(async (tx) => {
    const count = await tx.participationLog.count({ where: { scheduleId } });
    if (count >= schedule.activity.maxCapacity!) {
      throw ApiError.conflict('Activity is full');
    }
    await tx.participationLog.create({ data: { profileId, scheduleId } });
  });
```

Comment (gunnarwrld) [inline server/src/controllers/activities.controller.ts:1]: Both handlers type the Response as plain Response.

Fix both signatures to use the types:

  registerParticipation:
    `res: Response<ApiResponse<{ participantCount: number }>>`

  unregisterParticipation:
    `res: Response<ApiResponse<{ participantCount: number }>>`

File: server/src/db/createQueries.ts
Comment (gunnarwrld) [inline server/src/db/createQueries.ts:1]: Import paths missing .js extensions. PR #62 switched moduleResolution to NodeNext,  all internal imports need explicit extensions or TypeScript will fail to resolve them at runtime.

Change to:

  `import { Activity, FavoriteCreateDelete, ActivityDto } from "../types/index.js";`
  `import { prisma } from "./prisma.js";`

The types barrel is still the right import point the convention hasn't changed, only the path needs the .js suffix.

File: server/src/db/deleteQueries.ts
Comment (gunnarwrld) [inline server/src/db/deleteQueries.ts:1]: Same as `createQueries.ts`, import paths missing .js extensions.

 `import { prisma } from "./prisma.js";`
 `import { ActivityDto, FavoriteCreateDelete } from "../types/index.js";`

File: server/src/db/readQueries.ts [hunk 1]
```diff
@@ -203,4 +203,19 @@ export class READ {
...
+
+    static async participantCount(scheduleId: string): Promise<number> {
+        return prisma.participationLog.count({
+            where: { scheduleId }
+        })
+
+    }
+
+    static async isParticipating(profileId: string, scheduleId: string) {
+        return prisma.participationLog.findUnique({
```
gunnarwrld: `isParticipating` returns the full ParticipationLog row or null, but the controller only uses it as a truthiness check. 
Return `Promise<boolean>` instead to make the intent explicit and avoid leaking the internal model shape.

```
static async isParticipating(profileId: string, scheduleId: string): Promise<boolean> {
    const record = await prisma.participationLog.findUnique({
      where: { profileId_scheduleId: { profileId, scheduleId } }
    });
    return record !== null;
  }
```

gunnarwrld (changes requested):
Solid PR (as always lmao), logic is clean and the happy path covers the main cases well.

Also, something really minor:
- Missing newline at EOF in `deleteQueries.ts` and `activities.controller.ts`

server/src/controllers/activities.controller.ts
```diff
@@ -220,4 +220,61 @@ export const getActivities = asyncHandler(
       data,
     });
   }
-);
\ No newline at end of file
+);
+
+
+export const registerParticipation = asyncHandler(async (
```
JuriRappold, line 225: Same thing with the queries as in the other PR, *personally* id prefer if we keep all database interaction in the appropiate files.

JuriRappold (changes requested):

JuriRappold:
Nice job! Just the DB stuff, but thats more asthetics, than anything

server/src/controllers/activities.controller.ts
jasminerezai: You're right

JuriRappold (approved):
Noice job!!! Looks good to me!

server/src/db/createQueries.ts
```diff
@@ -65,4 +66,36 @@ export class CREATE{
...
+            const existing = await tx.participationLog.findUnique({
+                where: { profileId_scheduleId: { profileId, scheduleId } }
+            })
+            if (existing) throw ApiError.conflict('Already registered for this activity')
+
+            await tx.participationLog.create({
+                data: { profileId, scheduleId }
+            })
+
+            return tx.participationLog.count({ where: { scheduleId } })
```
gunnarwrld: This final count inside the transaction is not awaited:

This returns a `Promise<number>`, not a number. The transaction resolves correctly
because Prisma awaits the return value of the callback internally, but it's
inconsistent with every other query in this file and will confuse TypeScript
inference in stricter configs.

Just add await to it:

 `return await tx.participationLog.count({ where: { scheduleId } })`

server/src/controllers/activities.controller.ts
gunnarwrld: Good catch Juri

server/src/controllers/activities.controller.ts
```diff
@@ -3,7 +3,7 @@ import { asyncHandler } from '../middleware/asyncHandler.js';
 import { ApiError } from '../utils/ApiError.js';
 import { prisma, ProfileRole, ActivityStatus } from '../db/prisma.js';
 import { ApiResponse, UpdateScheduleStatusBody, UpdateScheduleStatusDto, Activity } from '../types/index.js';
-import { CreateActivitySchema, DeleteActivitySchema, UpdateActivityGeneralSchema, UpdateActivityURLSchema } from "../validators/index.js";
+import { CreateActivitySchema, DeleteActivitySchema, UpdateActivityGeneralSchema, UpdateActivityURLSchema, isUUID } from "../validators/index.js";
 import { DELETE, READ, UPDATE, CREATE } from "../db/queries.js";
 import { ActivityDto } from "../types/activity.types.js";
```
gunnarwrld: Here just import from the index barrel. This is a re-import in the same file, just add them to the `index,js` line above.

gunnarwrld (approved):
The enpoints and the hard problems from the first review are all fixed. Two things left and ready to go!

PR #62, opened 2026-05-15
chore(server): fix ESM module resolution for Railway deployment

gunnarwrld:
## What and Why

The server `tsconfig.json` was using `moduleResolution: "Bundler"` which works fine in local dev but is incompatible with how Railway runs the compiled output. Railway executes `node dist/server.js` directly (no bundler involved) and Node's native ESM loader requires all local import paths to include explicit `.js` file extensions. Without them, the server crashes at startup with `ERR_MODULE_NOT_FOUND`.

This was my mistake from when I originally suggested the `Bundler` setting.

## Changes

- Updated `tsconfig.json`: `"module"` and `"moduleResolution"` changed from `"ESNext"` / `"Bundler"` to `"NodeNext"` / `"NodeNext"`
- Added `.js` extensions to all internal import paths across the server source files

## Testing

Run locally to verify the build still works:

```bash
cd server
npm run build
node dist/server.js
```

`tsc` should compile with zero errors and `node dist/server.js` should start cleanly against a local `DATABASE_URL`.

## Related

Closes #33

Issue #33, opened 2026-05-04
Deployment


jasminerezai:
For the deployment I'd suggest:
Deploy backend to Railway, frontend to Vercel.

jasminerezai:
Gunnar, please start with this asap. I moved this to sprint 2 after today's meeting with Deema and Anton. We have moved around a bit, but focus on this. Thanks brother!

PR #65, opened 2026-05-16
fix(server): resolve Railway production build errors

gunnarwrld:
## What

Fixes a series of TypeScript errors that only surface during production build on Railway (`tsc` in strict mode without a bundler).

## Changes

**`server/package.json`**
- Updated build command to run `prisma generate --no-engine` before `tsc` so the generated Prisma types exist when the compiler runs
- Start command on Railway handles `prisma migrate deploy` at runtime (not during build) to avoid `DATABASE_URL` not being available at image build time

**`server/src/controllers/activities.controller.ts`**
- Fixed `TS2322`: `req.user.id` typed as `string | undefined`,  changed to `req.user!.id` since auth middleware guarantees the user exists at this point

**`server/src/db/readQueries.ts`**
- Fixed `TS7006`: added explicit types to `.map()` callback parameters in `activitiesFavoritedBy`, `participantsOf`, and `profilesFavorited` (implicit `any` was previously hidden by `moduleResolution: Bundler`)

## Why now

These errors were introduced when `moduleResolution` was changed from `Bundler` to `NodeNext` in PR #62 to support Railway deployment. `Bundler` mode is more lenient with implicit types; `NodeNext` with `strict: true` surfaces them.

## Related

Part of #33 (Deployment)

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "dev": "tsx watch src/server.ts",
-    "build": "tsc",
+    "build": "prisma generate --no-engine && tsc",
     "postbuild": "cp -r src/generated/prisma/. dist/generated/prisma/",
     "start": "node dist/server.js"
   },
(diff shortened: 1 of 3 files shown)


jasminerezai (approved):
Good Gunnar! All fixes are correct and necessary. The req.user!.id change is the right call, the explicit map types are proper TypeScript, and the build command change is the standard pattern for Prisma on Railway. Merge this.

JuriRappold (approved):
Same here, looks good to me!

PR #66, opened 2026-05-17
feat: added CRUD for schedule

### LINKED ISSUE #57: Schedule: create, update, and delete endpoints
POST /api/schedules --- restrict to `BOARD_MEMBER` and above
PUT/PATCH /api/schedules/:scheduleId --- restrict to `LEADER` and above, leader has to be of that activity
DELETE /api/schedules/:scheduleId --- restrict to `BOARD_MEMBER` and above

jasminerezai:
Added CRUD capabilities (Create, Update, Delete) to the `Schedule` resource, integrating appropriate role-based access control and leader activity ownership checks.

## Authorization and Security Rules
* **POST /api/schedules**: Restricted to `BOARD_MEMBER` and above.
* **PUT/PATCH /api/schedules/:scheduleId**: Restricted to `LEADER` and above.
  * Requesters with the `LEADER` role must be assigned as a leader of the activity associated with the schedule. `BOARD_MEMBER` and `ADMIN` skip this check.
  * Updating `activityId` is not supported (schedule-to-activity relationships are immutable). Only `startAt`, `endAt`, and `status` can be modified.
* **DELETE /api/schedules/:scheduleId**: Restricted to `BOARD_MEMBER` and above.

### Database Query Layer
* **createQueries.ts**: Implemented `CREATE.newSchedule` which inserts a new schedule and returns it formatted as a `ScheduleDto` containing activity and leader details.
* **updateQueries.ts**: Implemented `UPDATE.updateSchedule` to partially update schedules (`startAt`, `endAt`, `status`).
* **deleteQueries.ts**: Implemented `DELETE.deleteSchedule` to delete a schedule by its unique ID.

### Request Validation Layer (Zod)
* **schedule.validator.ts**: Added Zod validation schemas:
  * `CreateScheduleSchema`: Validates creation fields (`activityId`, `startAt`, `endAt`, `status`).
  * `UpdateScheduleSchema`: Validates updateable fields (`startAt`, `endAt`, `status`).
  * `ScheduleIdParamSchema`: Validates that `scheduleId` in path parameters is a valid UUID.
* **index.ts**: Added barrel exports for the new schemas.

### Controller & Routing Layer
* **schedule.controller.ts**:
  * Implemented `assertActivityAccess` to verify activity template existence and restrict non-assigned leaders from modifying schedules.
  * Added handlers `createSchedule`, `updateSchedule`, and `deleteSchedule`.
* **schedule.routes.ts**: Registered `POST /`, `PUT /:scheduleId`, `PATCH /:scheduleId`, and `DELETE /:scheduleId` with standard hierarchy role guards.

### DIFF
--- a/server/src/controllers/schedule.controller.ts
+++ b/server/src/controllers/schedule.controller.ts
@@ -6,6 +6,39 @@ import {ApiResponse} from "../types/index.js";
 import {ScheduleDto} from '../types/index.js';
 import {ScheduleDateSchema, ScheduleBoolWeekSchema} from "../validators/schedule.validator.js";
 import {output, ZodISODate, ZodLiteral, ZodNullable, ZodOptional, ZodUnion} from "zod";
+import { CREATE, UPDATE, DELETE } from "../db/queries.js";
+import { prisma, ProfileRole } from "../db/prisma.js";
+import { CreateScheduleSchema, UpdateScheduleSchema, ScheduleIdParamSchema } from "../validators/index.js";
+
+// Helper to assert activity existence & ownership for LEADERs
+async function assertActivityAccess(
+    activityId: string,
+    requesterId: string,
+    role: ProfileRole,
+) {
+    const activity = await prisma.activityTemplate.findUnique({
+        where: { id: activityId },
+        select: { id: true, name: true },
+    });
+
+    if (!activity) {
+        throw ApiError.notFound('Activity not found');
+    }
+
+    if (role === ProfileRole.LEADER) {
+        const ownership = await prisma.leaderActivity.findUnique({
+            where: {
+                profileId_activityId: { profileId: requesterId, activityId },
+            },
+        });
+
+        if (!ownership) {
+            throw ApiError.forbidden('You are not the leader of this activity');
+        }
+    }
+
+    return activity;
+}
 
 const currentWeek = asyncHandler(
     async (_req: Request, res: Response< ApiResponse< ScheduleDto[] > >) => {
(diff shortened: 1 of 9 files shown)


server/src/controllers/schedule.controller.ts
```diff
@@ -6,6 +6,39 @@ import {ApiResponse} from "../types/index.js";
...
+import { prisma, ProfileRole } from "../db/prisma.js";
+import { CreateScheduleSchema, UpdateScheduleSchema, ScheduleIdParamSchema } from "../validators/index.js";
+
+// Helper to assert activity existence & ownership for LEADERs
+async function assertActivityAccess(
+    activityId: string,
+    requesterId: string,
+    role: ProfileRole,
+) {
+    const activity = await prisma.activityTemplate.findUnique({
```
JuriRappold, line 19: Idk if its worth for such a query, but theoretically for a nice separation of concern i think it would be nice to put the queries into their query class. Result really isnt any different, so up to you.

JuriRappold: Also fairly sure `findUnique()` throws an error if it doesnt find a result

JuriRappold (changes requested):

JuriRappold:
Looks very good! Nice job! Just some things where you assert access rights

server/src/controllers/schedule.controller.ts
jasminerezai: findUnique() returns null when not found, it doesn't throw - the null check is needed. And agreed on the separation of concern, but will leave it for now since it doesn't affect the result. :))

JuriRappold (approved):
Looks good to me!

server/src/db/createQueries.ts
```diff
@@ -65,4 +65,47 @@ export class CREATE{
...
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        });
+
+        let leaders: any = schedule.activity.leaders ?? [];
```
gunnarwrld: The leaders formatting block is the same in both `newSchedule` and
`updateSchedule`:
```
  let leaders: any = schedule.activity.leaders ?? [];
  leaders = Array.isArray(leaders) ? leaders.map((el: { profile: any }) => el.profile) : [];
  (schedule.activity as any).leaders = undefined;
```
This is the same pattern that already exists in `readQueries.ts` (`anyWeekSchedule`).

Maybe it is a good idea to extract it into a shared utility in `utils/` or `db/`:

Then both `newSchedule` and `updateSchedule` call it. Reduces the mutation of
`schedule.activity.leaders` with (as any) casts in two separate files.

server/src/controllers/schedule.controller.ts
```diff
@@ -53,9 +86,112 @@ const getSchedule = asyncHandler(
...
+    let parsed;
+    try {
+        parsed = CreateScheduleSchema.parse(req.body);
+    } catch (err) {
+        throw ApiError.badRequest(`Invalid request body: ${err}`);
+    }
+
+    const { activityId, startAt, endAt, status } = parsed;
+
+    const activity = await prisma.activityTemplate.findUnique({
```
gunnarwrld: Here this fetch is redundant. Schedule creation is BOARD_MEMBER-only (no ownership restriction), so `assertActivityAccess` is intentionally not called here. But the manual existence check is unnecessary, if `activityId` doesn't exist, Prisma will throw a FK constraint error on create which the global error handler catches.

Remove the manual fetch and derive `finalStatus` directly from the schema default
or pass it from the client. Simplifies the handler to:
```
  const { activityId, startAt, endAt, status } = parsed;
  const newSchedule = await CREATE.newSchedule({ activityId, startAt, endAt, status });
```
If you need `defaultStatus `from the activity, move that logic into
`CREATE.newSchedule` itself so the controller stays thin.

gunnarwrld (approved):
Hey Jasmine! Everything looks good the structure is clean, validation is in place, role guards are correctly set up, and the query layer follows the existing patterns well. Just a few recommendations worth implementing before merging!

PR #67, opened 2026-05-17
feat: backend input validation

### LINKED ISSUE #24: Backend input validation
Validate all incoming request bodies using Zod (or express-validator).
Return clear 400 errors with field-level messages.
Cover: auth, activity creation/edit, participation, favorites.

I was thinking to put this in sprint 3, but I moved it up since I think it's good and it's better to validate early. Let me know if you want me to move it to next sprint instead.

JuriRappold:
Closes #24.
- Made ZodError handling uniform across input endpoints.
- Created a simpler error object.
- Added Zod Authentication for the auth route
- Added idSchema

## New Uniform Zod Error Handling
Can allow for multiple error causing lines of code to be in the try-catch block (DB queries for example) and better error distinction.
```typescript
let data: userInputLogin; // type infered by zod schema
  try {
    data = authLoginSchema.parse(req.body)
  } catch(error){
    if(error instanceof ZodError) throw ApiError.badRequest(JSON.stringify(parseZodError(error)))
    else throw ApiError.internal(`Something went wrong: ${error}`)
  }
```

```typescript
/*
Zod Schema:
export const authLoginSchema = z.object({
    email: z.email(),
    password: z.string().min(8),
});
*/

[
  {
    field: 'email',
    code: 'invalid_format',
    message: 'Invalid email address'
  },
  {
    field: 'password',
    code: 'too_small',
    message: 'Too small: expected string to have >=8 characters'
  }
]
```

- is an array of objects, where each object is an error cause by a bad input not adhering to the restrained (a bad input can throw multiple ZodErrors, if thats the case, the bad input is represented multiple times)
	- field --> property that failed the validation
	- code --> type of validation error
		- 'invalid_format' ex: `email: notEmail.com` --> returns 'invalid_format error'
		- 'too_small' ex: password has to be at least 8 characters
		- 'invalid_type' --> often returned when the input is undefined
	- message --> string message with additional information

**Using the received error obj:**
- Might have to explicitly cast the error string to json: `const errorObject = JSON.parse(res.body.error);`
- type checking the error would work as follows:
```typescript
error.forEach( err => {
        if( err.field === 'password'){
            if(err.code === 'too_small'){
                errMsg += "Password must be at least 8 characters!";
                errObj.errors.push( {field: err.field, type: err.code, message: "Password must be at least 8 characters!"})
            }
            if(err.code === 'invalid_type'){//means undefined in this case
                errMsg += "Please enter a Password";
            }
        }
        else if (err.field === 'email'){
            if(err.code === 'invalid_type'){//means undefined in this case
                errMsg += "Please enter an Email address";
            }
            if( err.code === 'invalid_format'){
                errMsg += "Please enter an Email address in the valid format: another@example.com"
            }
        }
    });
```

## Added Auth validation:

**Schema:**
```typescript
import z from 'zod';


export const authRegisterSchema = z.object({
    email: z.email(),
    password: z.string().min(8),
    name: z.nullish(z.string())
});


export const authLoginSchema = z.object({
    email: z.email(),
    password: z.string().min(8),
});
```

**Use Case:**
```typescript
export const registerHandler = asyncHandler(async (
  req: Request,
  res: Response<ApiResponse<AuthResponseDto>>,
) => {
  let newUser: userInputRegister;
  try{
    newUser = authRegisterSchema.parse(req.body);
  } catch (error){
    if(error instanceof ZodError){
      throw ApiError.badRequest(JSON.stringify(parseZodError(error)))
    }
    else{
      throw ApiError.internal(`Something went wrong: ${error}`)
    }
  }
  const result: AuthResponseDto = await register(newUser)
  res.status(201).json({
    status: 'success',
    data: result,
  });
});

export const loginHandler = asyncHandler(async (
  req: Request,
  res: Response<ApiResponse<AuthResponseDto>>,
) => {
  let data: userInputLogin;
  try {
    data = authLoginSchema.parse(req.body)
  } catch(error){
    if(error instanceof ZodError) throw ApiError.badRequest(JSON.stringify(parseZodError(error)))
    else throw ApiError.internal(`Something went wrong: ${error}`)
  }

    // const {email, password} = req.body;
    const result = await login(data.email, data.password);

    res.status(200).json({
      status: 'success',
      data: result,
    });
  // }
});
```

## Added idSchema
Added this schema, so any id can be checked, regardless of where the id is from (noticed multiple validation files had essentially this)

```typescript
import z from 'zod';

export const IdSchema = z.uuid();

export function isUUID(id: string): boolean
{
    return IdSchema.safeParse(id).success;
}
```

### DIFF
--- a/server/src/controllers/activities.controller.ts
+++ b/server/src/controllers/activities.controller.ts
@@ -3,9 +3,16 @@ import { asyncHandler } from '../middleware/asyncHandler.js';
 import { ApiError } from '../utils/ApiError.js';
 import { prisma, ProfileRole, ActivityStatus } from '../db/prisma.js';
 import { ApiResponse, UpdateScheduleStatusBody, UpdateScheduleStatusDto, Activity } from '../types/index.js';
-import { CreateActivitySchema, DeleteActivitySchema, UpdateActivityGeneralSchema, UpdateActivityURLSchema } from "../validators/index.js";
+import {
+    CreateActivitySchema,
+    DeleteActivitySchema,
+    parseZodError,
+    UpdateActivityGeneralSchema,
+    UpdateActivityURLSchema
+} from "../validators/index.js";
 import { DELETE, READ, UPDATE, CREATE } from "../db/queries.js";
 import { ActivityDto } from "../types/activity.types.js";
+import {ZodError} from "zod";
 
 // ──────────────────────────────────────────────────────────────
 // Shared helpers
@@ -161,7 +168,8 @@ export const updateActivity = asyncHandler(
       updateBody = UpdateActivityGeneralSchema.parse(req.body);
     } catch (error) {
       console.error(error);
-      throw ApiError.badRequest(`Invalid request body or params: ${error}`);
+      if(error instanceof ZodError) throw ApiError.badRequest(`Invalid request body or params: ${parseZodError(error)}`);
+      else throw ApiError.internal(`Something went wrong: ${error}`)
     }
 
     // Check if activity exists before attempting update
--- a/server/src/controllers/activities.controller.ts
+++ b/server/src/controllers/activities.controller.ts
@@ -2,16 +2,16 @@ import { Request, Response } from 'express';
 import { asyncHandler } from '../middleware/asyncHandler.js';
 import { ApiError } from '../utils/ApiError.js';
 import { prisma, ProfileRole, ActivityStatus } from '../db/prisma.js';
-import { ApiResponse, UpdateScheduleStatusBody, UpdateScheduleStatusDto, Activity } from '../types/index.js';
+import { ApiResponse, UpdateScheduleStatusBody, UpdateScheduleStatusDto, Activity, ActivityDto } from '../types/index.js';
 import {
     CreateActivitySchema,
     DeleteActivitySchema,
     parseZodError,
     UpdateActivityGeneralSchema,
-    UpdateActivityURLSchema
+    UpdateActivityURLSchema,
+    StatusValidationSchema, IdSchema,
 } from "../validators/index.js";
 import { DELETE, READ, UPDATE, CREATE } from "../db/queries.js";
-import { ActivityDto } from "../types/activity.types.js";
 import {ZodError} from "zod";
 
 // ──────────────────────────────────────────────────────────────
@@ -76,16 +76,18 @@ export const updateScheduleStatusHandler = asyncHandler(async (
   req: Request<{ activityId: string; scheduleId: string; }, any, UpdateScheduleStatusBody>,
   res: Response<ApiResponse<UpdateScheduleStatusDto>>,
 ) => {
-  const { activityId, scheduleId } = req.params;
+  let activityId: string;
+  let scheduleId: string;
+  let status: ActivityStatus;
   const { id: requesterId, role } = req.user!;
-  const { status } = req.body;
+  try{
+      activityId = IdSchema.parse(req.params.activityId);
+      scheduleId = IdSchema.parse(req.params.scheduleId);
+      status = StatusValidationSchema.parse(req.body);
 
-  // ── 1. Validate the incoming status value ────────────────────
-  const validStatuses = Object.values(ActivityStatus) as string[];
-  if (!status || !validStatuses.includes(status)) {
-    throw ApiError.badRequest(
-      `Invalid status. Must be one of: ${validStatuses.join(', ')}`,
-    );
+  } catch (error){
+      if( error instanceof ZodError) throw ApiError.badRequest(JSON.stringify(parseZodError(error)));
+      else throw ApiError.internal(`Something went wrong: ${error}`)
   }
 
   // ── 2. Activity existence + ownership ────────────────────────
@@ -139,7 +141,6 @@ export const newActivity = asyncHandler(
     try {
       newActivity = CreateActivitySchema.parse(req.body);
     } catch (error) {
-      console.error(error);
       throw ApiError.badRequest(`Invalid request body: ${error}`);
     }
 
@@ -167,7 +168,6 @@ export const updateActivity = asyncHandler(
       updateParams = UpdateActivityURLSchema.parse(req.params);
       updateBody = UpdateActivityGeneralSchema.parse(req.body);
     } catch (error) {
-      console.error(error);
       if(error instanceof ZodError) throw ApiError.badRequest(`Invalid request body or params: ${parseZodError(error)}`);
       else throw ApiError.internal(`Something went wrong: ${error}`)
     }
@@ -202,7 +202,6 @@ export const deleteActivity = asyncHandler(
     try {
       deleteParams = DeleteActivitySchema.parse(req.params);
     } catch (error) {
-      console.error(error);
       throw ApiError.badRequest(`Invalid request params: ${error}`);
     }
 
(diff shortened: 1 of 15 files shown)


server/src/controllers/user.controller.ts
```diff
@@ -18,7 +19,8 @@ const newFavorites = asyncHandler(
     try{
         ids = CreateFavoriteSchema.parse({profileId: req.user.id, activityId: req.params.activityId})
     } catch(error){
-        throw ApiError.badRequest(`Invalid ids: ${error}`)
+        if(error instanceof ZodError) throw ApiError.badRequest(`Invalid ids: ${parseZodError(error)}`)
```
jasminerezai: parseZodError() returns an array of objects. Using it inside a template literal like ${parseZodError(error)} will coerce it to [object Object] instead of the actual error data.  Change to if(error instanceof ZodError) throw ApiError.badRequest(JSON.stringify(parseZodError(error)))

server/src/controllers/auth.controller.ts
```diff
@@ -39,18 +40,22 @@ export const loginHandler = asyncHandler(async (
...
 ) => {
-  const { email, password } = req.body;
-
-  if (!email || !password) {
-    throw ApiError.badRequest('Please provide email and password');
+  let data: userInputLogin;
+  try {
+    data = authLoginSchema.parse(req.body)
+  } catch(error){
+    if(error instanceof ZodError) throw ApiError.badRequest(JSON.stringify(parseZodError(error)))
```
jasminerezai: JSON.stringify(parseZodError(error)) sends a raw JSON array string as the error message. The frontend (LoginPage.jsx, RegisterPage.jsx) renders err.message directly in the error banner so users will see [{"field":"email","code":"invalid_format",...}] on screen. Change to plain string:

if(error instanceof ZodError) throw ApiError.badRequest('Invalid email or password format.')

server/src/controllers/auth.controller.ts
```diff
@@ -39,18 +40,22 @@ export const loginHandler = asyncHandler(async (
...
+    else throw ApiError.internal(`Something went wrong: ${error}`)
   }
 
-  const result = await login(email, password);
+    // const {email, password} = req.body;
```
jasminerezai: commented-out code, please remove.

jasminerezai (changes requested):
Good job! 3 things to fix before this can merge

JuriRappold:
Implemented the changes. Now just waiting for @gunnarwrld :)

server/src/types/auth.types.ts
```diff
@@ -1,4 +1,8 @@
 import { ProfileRole } from '../db/prisma.js';
+import {authLoginSchema, authRegisterSchema} from "../validators/index.js";
```
gunnarwrld: The types layer should never import from the validators layer. Types are the
bottom of the dependency graph, validators import types, not the other way around.

Move the inferred types into the validator file itself and re-export from
the barrel, or inline the `z.infer<>`  in the validator and import from there
in the controller directly:
```
  // auth.validation.ts
  export const authRegisterSchema = z.object({ ... });
  export type UserInputRegister = z.infer<typeof authRegisterSchema>;
  export type UserInputLogin = z.infer<typeof authLoginSchema>;

  // types/index.ts barrel re-exports them
  export type { UserInputRegister, UserInputLogin } from '../validators/auth.validation.js';
```

server/src/types/zod.types.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * A bit simpler than an entire ZodError.
+ * NOTE THAT THE FIRST LETTER IS A SMALL Z!!!!
+ * ONLY DIFFERENCE TO THE ORIGINAL ZodError
+ */
+export interface zodError {
+    field: string,
+    code: string,
+    message: string
+}
```
gunnarwrld, line 10: `zodError` is used in `zodError.formatting.ts` but the interface itself is never
exported from `types/index.ts.` Any future file that wants to type the `parseZodError`
return value has to import directly from the individual file, breaking the barrel
convention. Add it to `types/index.ts`:

  `export type { zodError } from './zod.types.js';`

server/src/controllers/auth.controller.ts
```diff
@@ -16,19 +18,18 @@ export const registerHandler = asyncHandler(async (
...
-
-  // Basic validation
-  if (password.length < 8) {
-    throw ApiError.badRequest('Password must be at least 8 characters long');
+  let newUser: userInputRegister;
+  try{
+    newUser = authRegisterSchema.parse(req.body);
+  } catch (error){
+    if(error instanceof ZodError){
+      throw ApiError.badRequest('Invalid email or password format.')
```
gunnarwrld, line 26: Here `registerHandler` throws a generic message on `ZodError`, but `loginHandler` and every other handler in the codebase throws the structured `parseZodError` output:

  `throw ApiError.badRequest(JSON.stringify(parseZodError(error)))`

The frontend relies on the structured error object to display field-level messages. `registerHandler` breaks that contract for the register form. Make it consistent:
```
  if (error instanceof ZodError) {
    throw ApiError.badRequest(JSON.stringify(parseZodError(error)))
  }
```

server/src/controllers/activities.controller.ts
```diff
@@ -161,7 +168,8 @@ export const updateActivity = asyncHandler(
       updateBody = UpdateActivityGeneralSchema.parse(req.body);
     } catch (error) {
       console.error(error);
```
gunnarwrld: This was already present before this PR but is now more visible. `console.error`
in a production handler leaks stack traces and internal error details to logs.
Remove it,  the error is already forwarded to ApiError which the global
error handler processes.

gunnarwrld (changes requested):
The circular dependency in `auth.types.ts` blocks merge, types importing from validators inverts the dependency graph. The commented-out files (not good to commit empty ones) and the inconsistent error message in registerHandler also need cleaning up before this goes in.

server/src/controllers/auth.controller.ts
JuriRappold: Yeah i had that, but then Jasmin informed me that the object would be directly displayed to the user (one of the resolved conversations). So I added the generic error message

server/src/types/auth.types.ts
JuriRappold: Good Catch!

server/src/controllers/auth.controller.ts
jasminerezai: The frontend (LoginPage.jsx, RegisterPage.jsx) renders err.message directly in the error banner - sending JSON.stringify(parseZodError(error)) would show raw JSON to the user. Plain string is intentional here.

jasminerezai: see above, plain string is correct for auth. But Gunnar's point on auth.types.ts is valid, you need to move the inferred types into auth.validation.ts and re-export from the barrel. Also you have merge conflicts with main now in activities.controller.ts, schedule.controller.ts and validators/index.ts. Resolve those and it should be good to go.

JuriRappold: Definetly! Already moved them!

jasminerezai (approved):

server/src/types/index.ts
```diff
@@ -30,7 +30,17 @@ export interface PaginationQuery {
...
 // ── Domain types ────────────────────────────────────────────────
 
-export type { UserDto, AuthResponseDto, MeResponseDto } from './auth.types.js';
+export type {
+  UserDto,
+  AuthResponseDto,
+  MeResponseDto,
+} from './auth.types.js';
+
+export{
```
gunnarwrld: These are types, not values. Missing the `type` keyword will cause issues in
stricter TypeScript configs and is inconsistent with every other re-export in
this file which uses `export type { ... }`.

gunnarwrld (approved):
Great! ready to merge, just check my last comment.

server/src/types/index.ts
JuriRappold:
```suggestion
export type {
```

PR #68, opened 2026-05-18
implemented GET endpoint for profiles

### LINKED ISSUE #54: Profile page -- backend endpoint
Create a protected API endpoint for the profile page.

- GET /api/profile (or /api/users/me)
- Returns logged-in user's name, email and role
- Returns their upcoming registered activities
- Returns their saved favorite sports/activities
- Must be protected (requires valid token/session)

JuriRappold:
Closes #54.
- **Only implements the route in the description of the issue**
-  No valid id check --> we get the users id through: `req.user.id` which guarantees a valid id and a logged in user. Uses the auth-middleware also used by the favorites-endpoints
- Route: **GET** `/api/users/me`
- DTO used in respons *ProfileDto* (uses ActivityDto and ScheduleDto)
- Example response object: `{status: "success", date: ProfileDto}`

### ProfileDto
```typescript
import {ScheduleDto, ActivityDto} from "./index.js";
import { Profile } from '../generated/prisma/index.js';

export type ProfileDto = Pick<Profile, 'profileName' | 'email' | 'role'> & {
    favorites: ActivityDto[],
    participations: ScheduleDto[]
}
```

- to get favorites or participations: destructure the object:
	- `const {favorites} = fullProfile.data;`
	- `const {participations} = fullProfile.data;`

### Example Return Object:
```JSON
{
  "status": "success",
  "data": {
    "email": "user4@example.com",
    "profileName": null,
    "role": "LEADER",
    "favorites": [
      {
        "id": "40cc5b24-4761-42ea-9be5-3a36c262658b",
        "name": "Swimming",
        "description": "Swimming session",
        "location": "Location 4",
        "maxCapacity": 16,
        "defaultStatus": "ACTIVE",
        "notes": null,
        "createdAt": "2026-05-11T16:12:24.158Z",
        "updatedAt": "2026-05-11T16:12:24.158Z",
        "leaders": [
          {
            "profileId": "49f33b55-8bb9-4700-839a-0f9209ab4c4b",
            "activityId": "40cc5b24-4761-42ea-9be5-3a36c262658b"
          }
        ],
        "timeSlots": [
          {
            "id": "4ca4a51b-2f08-4fc9-817c-a0435b9b9549",
            "activityId": "40cc5b24-4761-42ea-9be5-3a36c262658b",
            "weekday": "FRIDAY",
            "startTime": "1970-01-01T17:00:00.000Z",
            "endTime": "1970-01-01T18:00:00.000Z"
          }
        ]
      }
    ],
    "participations": [
      {
        "id": "2f72d41e-6f97-4991-8433-d62233234516",
        "activityId": "40cc5b24-4761-42ea-9be5-3a36c262658b",
        "createdAt": "2026-05-11T16:12:24.322Z",
        "updatedAt": "2026-05-11T16:12:24.322Z",
        "startAt": "2026-05-22T16:00:00.275Z",
        "endAt": "2026-05-22T17:00:00.275Z",
        "status": "ACTIVE",
        "activity": {
          "id": "40cc5b24-4761-42ea-9be5-3a36c262658b",
          "name": "Swimming",
          "description": "Swimming session",
          "location": "Location 4",
          "maxCapacity": 16,
          "defaultStatus": "ACTIVE",
          "notes": null,
          "createdAt": "2026-05-11T16:12:24.158Z",
          "updatedAt": "2026-05-11T16:12:24.158Z"
        },
        "leaders": [
          {
            "id": "49f33b55-8bb9-4700-839a-0f9209ab4c4b",
            "profileName": null
          }
        ]
      },
      {
        "id": "d499f40e-9065-441b-88fb-a07894c43e74",
        "activityId": "7898d9bc-59ad-4978-aa55-6efca65d16b6",
        "createdAt": "2026-05-11T16:12:24.346Z",
        "updatedAt": "2026-05-11T16:12:24.346Z",
        "startAt": "2026-05-26T08:00:00.275Z",
        "endAt": "2026-05-26T09:00:00.275Z",
        "status": "ACTIVE",
        "activity": {
          "id": "7898d9bc-59ad-4978-aa55-6efca65d16b6",
          "name": "Pilates",
          "description": "Pilates session",
          "location": "Location 7",
          "maxCapacity": 22,
          "defaultStatus": "ACTIVE",
          "notes": null,
          "createdAt": "2026-05-11T16:12:24.161Z",
          "updatedAt": "2026-05-11T16:12:24.161Z"
        },
        "leaders": [
          {
            "id": "d444bdfe-62c6-46d8-a627-d0ebbaa8e911",
            "profileName": null
          }
        ]
      },
      {
        "id": "c4bb9022-27ac-4af2-951a-6db54d99e41d",
        "activityId": "0cec1395-c75d-4b4a-823e-d21a6e8bdb84",
        "createdAt": "2026-05-11T16:12:24.358Z",
        "updatedAt": "2026-05-11T16:12:24.358Z",
        "startAt": "2026-05-18T08:00:00.275Z",
        "endAt": "2026-05-18T09:00:00.275Z",
        "status": "ACTIVE",
        "activity": {
          "id": "0cec1395-c75d-4b4a-823e-d21a6e8bdb84",
          "name": "Boxing",
          "description": "Boxing session",
          "location": "Location 10",
          "maxCapacity": 28,
          "defaultStatus": "ACTIVE",
          "notes": null,
          "createdAt": "2026-05-11T16:12:24.164Z",
          "updatedAt": "2026-05-11T16:12:24.164Z"
        },
        "leaders": [
          {
            "id": "d444bdfe-62c6-46d8-a627-d0ebbaa8e911",
            "profileName": null
          }
        ]
      }
    ]
  }
}
```

### DIFF
--- a/server/src/controllers/user.controller.ts
+++ b/server/src/controllers/user.controller.ts
@@ -2,8 +2,9 @@ import {ApiError} from "../utils/ApiError.js";
 import {Request, Response} from "express";
 import {asyncHandler} from "../middleware/asyncHandler.js";
 import {READ, CREATE, DELETE} from '../db/queries.js';
-import {CreateFavoriteSchema, DeleteFavoriteSchema} from "../validators/index.js";
-import {ApiResponse, FavoriteCreateDelete, ActivityDto} from "../types/index.js";
+import { CreateFavoriteSchema, DeleteFavoriteSchema } from "../validators/index.js";
+import { ApiResponse, FavoriteCreateDelete, ActivityDto, ProfileDto } from "../types/index.js";
+import { IdSchema } from "../validators/profile.validator.js";
 
 const getFavorites = asyncHandler(
     async (req: Request, res: Response<ApiResponse<ActivityDto[]>>) => {
@@ -38,8 +39,32 @@ const deleteFavorites = asyncHandler(
     res.status(204).send()
 });
 
+const getFullProfile = asyncHandler(
+    async (req: Request, res: Response<ApiResponse<ProfileDto> > ) => {
+        const isValidId = IdSchema.safeParse(req.user.id);
+        if(!isValidId.success){
+            throw ApiError.badRequest(JSON.stringify(isValidId.error))
+        }
+        else{
+            let fullProfile: ProfileDto;
+            try{
+                fullProfile = await READ.fullProfile(req.user.id);
+            } catch (error){
+                throw ApiError.internal(`Something went wrong: ${error}`)
+            }
+            if(fullProfile){
+                res.status(200).json({status: "success", data: fullProfile});
+            }
+            else{
+                throw ApiError.badRequest(`Couldn't find the profile: ${req.user.id}`)
+            }
+        }
+    }
+)
+
 export const controller = {
     getFavorites,
     newFavorites,
     deleteFavorites,
+    getFullProfile,
 }
--- a/server/src/controllers/user.controller.ts
+++ b/server/src/controllers/user.controller.ts
@@ -4,7 +4,6 @@ import {asyncHandler} from "../middleware/asyncHandler.js";
 import {READ, CREATE, DELETE} from '../db/queries.js';
 import { CreateFavoriteSchema, DeleteFavoriteSchema } from "../validators/index.js";
 import { ApiResponse, FavoriteCreateDelete, ActivityDto, ProfileDto } from "../types/index.js";
-import { IdSchema } from "../validators/profile.validator.js";
 
 const getFavorites = asyncHandler(
     async (req: Request, res: Response<ApiResponse<ActivityDto[]>>) => {
@@ -41,23 +40,17 @@ const deleteFavorites = asyncHandler(
 
 const getFullProfile = asyncHandler(
     async (req: Request, res: Response<ApiResponse<ProfileDto> > ) => {
-        const isValidId = IdSchema.safeParse(req.user.id);
-        if(!isValidId.success){
-            throw ApiError.badRequest(JSON.stringify(isValidId.error))
+        let fullProfile: ProfileDto;
+        try{
+            fullProfile = await READ.fullProfile(req.user.id);
+        } catch (error){
+            throw ApiError.internal(`Something went wrong: ${error}`)
+        }
+        if(fullProfile){
+            res.status(200).json({status: "success", data: fullProfile});
         }
         else{
-            let fullProfile: ProfileDto;
-            try{
-                fullProfile = await READ.fullProfile(req.user.id);
-            } catch (error){
-                throw ApiError.internal(`Something went wrong: ${error}`)
-            }
-            if(fullProfile){
-                res.status(200).json({status: "success", data: fullProfile});
-            }
-            else{
-                throw ApiError.badRequest(`Couldn't find the profile: ${req.user.id}`)
-            }
+            throw ApiError.badRequest(`Couldn't find the profile: ${req.user.id}`)
         }
     }
 )
--- a/server/src/controllers/user.controller.ts
+++ b/server/src/controllers/user.controller.ts
@@ -40,18 +40,8 @@ const deleteFavorites = asyncHandler(
 
 const getFullProfile = asyncHandler(
     async (req: Request, res: Response<ApiResponse<ProfileDto> > ) => {
-        let fullProfile: ProfileDto;
-        try{
-            fullProfile = await READ.fullProfile(req.user.id);
-        } catch (error){
-            throw ApiError.internal(`Something went wrong: ${error}`)
-        }
-        if(fullProfile){
-            res.status(200).json({status: "success", data: fullProfile});
-        }
-        else{
-            throw ApiError.badRequest(`Couldn't find the profile: ${req.user.id}`)
-        }
+        const fullProfile: ProfileDto = await READ.fullProfile(req.user.id);
+        res.status(200).json({status: "success", data: fullProfile});
     }
 )
 
--- a/server/src/db/readQueries.ts
+++ b/server/src/db/readQueries.ts
@@ -1,7 +1,8 @@
 import { prisma } from "./prisma.js";
 import { startAndEndOfWeek } from "../utils/weekCalculator.js";
 import { ActivityTemplate, Profile } from "../generated/prisma/index.js";
-import { ScheduleDto, ActivityDto } from '../types/index.js';
+import {ScheduleDto, ActivityDto, ProfileDto} from '../types/index.js';
+import {ApiError} from "../utils/ApiError.js";
 export class READ {
     /**
  * returns user based of their unique email
@@ -113,7 +114,7 @@ export class READ {
 
     /**
      * @param profileId --> requires to be logged-in, i.e. we need an account
-     * @return Promise<ActivityTemplate[]>
+     * @return Promise<ActivityDto[]>
      *     **FAIL**
      *     --> is empty if the profile doesn't have favorite activities
      *     **SUCCESS**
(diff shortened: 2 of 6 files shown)


JuriRappold:
Note that I also have the DELETE and PATCH route for this endpoint ready, but not yet committed, as neither route was part of the issue description.

server/src/validators/profile.validator.ts
gunnarwrld, line 1: This file defines:
```
  export const IdSchema = z.uuid();
```
PR #67 (already in flight) adds the exact same export in
`idSchema.validator.ts` and exports it from the validators barrel as `IdSchema`.

This `profile.validator.ts` is not exported from `validators/index.ts`, so
`user.controller.ts` imports it directly from the individual file:

This breaks the barrel convention and creates a second source of truth for the
same utility. Once PR #67 merges, delete `profile.validator.ts` entirely and
change the import to the barrel:
```
  import { IdSchema } from "../validators/index.js";
```

server/src/db/readQueries.ts
```diff
@@ -203,4 +204,106 @@ export class READ {
...
+                                                profileName: true
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+                        }
+     */
+
+    static async fullProfile(activityId: string): Promise<ProfileDto> {
```
gunnarwrld: `fullProfile` calls two separate Prisma queries sequentially after the profile
fetch:
``
  const favorites = await this.activitiesFavoritedBy(profile.id);
  const participation = await this.activitiesParticipatedBy(profile.id);
``
Each of those is its own query, meaning a single `GET /me` triggers 3 round
trips to the database. Collapse them into the initial profile fetch using
include:
```
  const profile = await prisma.profile.findUnique({
      where: { id: profileId },
      select: {
          id: true,
          profileName: true,
          email: true,
          role: true,
          favorites: { include: { activity: { include: { ... } } } },
          participationLogs: { include: { schedule: { include: { ... } } } }
      }
  })
```
Then format in one pass. Eliminates 2 of the 3 DB round trips.

gunnarwrld: Also, the method signature is:

  `static async fullProfile(activityId: string): Promise<ProfileDto>`

The parameter is named activityId but it's used as a profileId:

  `where: { id: activityId }`

This is an error from the surrounding file context. Rename it:
```
  static async fullProfile(profileId: string): Promise<ProfileDto>
```

server/src/controllers/user.controller.ts
```diff
@@ -38,8 +39,32 @@ const deleteFavorites = asyncHandler(
     res.status(204).send()
 });
 
+const getFullProfile = asyncHandler(
+    async (req: Request, res: Response<ApiResponse<ProfileDto> > ) => {
+        const isValidId = IdSchema.safeParse(req.user.id);
```
gunnarwrld: This handler validates `req.user.id` against `IdSchema`:

req.user is set by `authMiddleware` from the verified JWT payload. The id in
the token was set at registration time from a Prisma-generated UUID. If it's
not a valid UUID, that is a server-side integrity failure, not a bad client
request, it should be a 500, not a 400. And since this is `/me` (no
client-supplied id), this validation adds no real security value.

Remove the UUID check entirely. The guard is `authMiddleware` itself.
If `fullProfile` returns null, that's already handled by `READ.fullProfile`
throwing `ApiError.badRequest`. Simplify to:
```
  const fullProfile = await READ.fullProfile(req.user.id);
  res.status(200).json({ status: 'success', data: fullProfile });
```

server/src/types/profile.types.ts
```diff
@@ -0,0 +1,7 @@
+import {ScheduleDto, ActivityDto} from "./index.js";
```
gunnarwrld: `profile.types.ts` imports from the barrel and `index.ts` re-exports `ProfileDto` from `profile.types.ts`. This creates a circular import within the types layer.

gunnarwrld (changes requested):
Solve this issues and just make sure this branch rebases on top of #67 and removes `profile.validator.ts`.

jasminerezai (approved):

JuriRappold:
Got the merge conflict covered

server/src/db/readQueries.ts
```diff
@@ -203,4 +205,124 @@ export class READ {
...
+                        }
+                    }
+                }
+            }
+        })
+        if(!profile){
+            throw ApiError.badRequest("Invalid Id")
+        }
+        else{
+            const favorites: ActivityDto[] = await this.activitiesFavoritedBy(profile.id);
```
gunnarwrld: The single `findUnique` with includes fetches the raw data correctly, but then
`fullProfile` still calls two separate methods after it:

The profile fetch already has favorites and participationLogs included, that data is sitting in the profile object unused. Format it inline from `profile.favorites` and `profile.participationLogs` instead of firing two more queries. The `activitiesFavoritedBy` and `activitiesParticipatedBy` calls are now
redundant.

server/src/controllers/user.controller.ts
```diff
@@ -38,8 +38,26 @@ const deleteFavorites = asyncHandler(
     res.status(204).send()
 });
 
+const getFullProfile = asyncHandler(
+    async (req: Request, res: Response<ApiResponse<ProfileDto> > ) => {
+        let fullProfile: ProfileDto;
+        try{
+            fullProfile = await READ.fullProfile(req.user.id);
```
gunnarwrld: The `try/catch` wraps `READ.fullProfile` and rethrows everything as
`ApiError.internal`. But `READ.fullProfile` throws `ApiError.badRequest` when the
profile is not found. Wrapping it in a blanket catch and rethrowing as
internal swallows that specific error and turns a 400 into a 500.

Since `asyncHandler` already forwards unhandled errors to the global error
handler, the `try/catch` is unnecessary here. Remove it entirely:
```
  const getFullProfile = asyncHandler(async (req, res) => {
      const fullProfile = await READ.fullProfile(req.user.id);
      res.status(200).json({ status: 'success',  fullProfile });
  })

```

`READ.fullProfile` throws `ApiError.badRequest` if not found, which propagates correctly through `asyncHandler` to the error handler. The null check after
the call is also redundant since `fullProfile` never returns null, it either
returns a `ProfileDto` or throws.

server/src/routes/user.routes.ts
```diff
@@ -16,9 +16,19 @@ GET /users/me/favorites — return user's favorite sports list
...
+
+/*
+PROFILE
+----------
+GET ./api/users/me -- returns full profile (email, profileName, favorites: ActivityDto[], participations: ScheduleDto[] )
+
+PATCH ./api/users/me -- update email, profileName, or role?? NOT IN ISSUE DESCRIPTION
+ */
+
+userRoutes.get('/me', userController.getFullProfile)
```
gunnarwrld, line 31: `GET /me`  is registered without  `authMiddleware` . Since it reads  `req.user.id` , hitting it without a token it will throw a runtime error. Add  `authMiddleware`

gunnarwrld (changes requested):
The missing  `authMiddleware`  on  `GET /me`  will cause a runtime crash. The N+1 fix is half done, the includes are there but the two extra queries are still fired. And the blanket  `ApiError.internal`  in the controller swallows legitimate 400's from the query layer.

server/src/routes/user.routes.ts
JuriRappold: But wouldn't this at the top of the file be enough: `userRoutes.use(authMiddleware)`?

gunnarwrld: YESSS, sorry i just had the changes open and forgot about that line. It is enough

server/src/db/readQueries.ts
```diff
@@ -276,42 +260,65 @@ export class READ {
...
+
             }
         })
         if(!profile){
             throw ApiError.badRequest("Invalid Id")
         }
         else{
-            const favorites: ActivityDto[] = await this.activitiesFavoritedBy(profile.id);
-            const participation: ScheduleDto[] = await this.activitiesParticipatedBy(profile.id);
+            const favorites: ActivityDto[] = profile.favorites.map(el => el.activity);
```
gunnarwrld, line 281: `profile.favorites` comes from the Favorite join table with include: `{ activity: {...} }`. `el.activity` here includes `timeSlots` and `leaders` as raw join rows (`leaderActivity shape: { profileId, activityId }`), not the formatted leaders array that `ActivityDto` expects. 

The `leaders` field in `ActivityDto` should be `{ id, profileName }[]` but the raw Prisma include returns
`{ profileId: string, activityId: string }[]`.

Verify that the `ActivityDto` type actually accepts the raw `leaderActivity` shape, or apply the same leaders formatting used for participations.

gunnarwrld (approved):

server/src/db/readQueries.ts
JuriRappold: the `{id, profileName }` is for ScheduleDto, not for ActivityDto. But I get the confusion, its not consistent in the DTOs, my bad.
**ScheduleDto:**
```typescript
export type ScheduleDto = Schedule &{
    activity: ActivityTemplate,
    leaders: Pick<Profile, 'id' | 'profileName'>[]
};
```
**ActivityDto:**
```typescript
export type Activity = {
  name: string,
  location: string,
  leaders: string[],
  description?: string | null,
  notes?: string | null,
  maxCapacity?: number | null,
  defaultStatus?: ActivityStatus,
  timeSlots: TimeSlot[],
};
export interface ActivityDto extends Omit<Activity, 'timeSlots' | 'leaders'> {
  leaders: {
    profileId: string;
    activityId: string;
  }[];
  timeSlots: {
    id: string;
    activityId: string;
    weekday: Weekday;
    startTime: Date;
    endTime: Date;
  }[];
}
```

JuriRappold: so it does adhere to the `ActivityDto` shape, but the exact shape of leaders differ from Dto to Dto. Which is an inconsistency I should've caught

jasminerezai (approved):
Good job Juri

server/src/db/readQueries.ts
gunnarwrld: Ohhh yeah that makes sense, and thanks for the clarification! just noticed that `ActivityDto.leaders` does stick to its own type definition, my bad for flagging it as a bug. That said, the inconsistency between the two DTOs is a real issue worth tracking. `ActivityDto` exposes the raw join table shape `{ profileId, activityId }` while `ScheduleDto` uses `Pick<Profile, 'id' | 'profileName'>[]`, any frontend component consuming both will have to handle leaders differently depending on context, which is going to cause confusion especially as we build out the conditional rendering work. i'll open an issue to track normalizing both to `Pick<Profile, 'id' | 'profileName'>[]` as a follow-up.

PR #69, opened 2026-05-19
Add filter controls to schedule page (sport, day, favorites)

### LINKED ISSUE #27: Schedule filters -- sport, day, time
Add filter controls to the schedule page:
- Filter by sport type
- Filter by date / day of week
- Filter by favorite sports (if logged in)

larrymijo:
## What was implemented
Added a filter bar above the calendar on the schedule page with three filters:

- **Sport** — dropdown auto-populated from the activities currently loaded
- **Day of week** — Any day / Monday / Tuesday / ... / Sunday
- **Favorites only** — toggle that only appears when the user is logged in, filters the schedule down to the user's favorited sports
- Plus a **Clear filters** button that resets all three at once

## Approach
- New presentational component `client/src/components/ScheduleFilters.jsx`. It just renders the controls and bubbles changes back up via props — no state or fetching inside it, so we can reuse it later (e.g. on ActivitiesPage).
- Filter state lives in `SchedulePage.jsx`. Sport + favorites are applied to a memoized `filteredActivities` list. Day-of-week is applied inside `getActivitiesForDay` since it depends on the specific date of each calendar cell.
- Favorites are fetched from `/api/users/me/favorites` using the existing `FavoritesService.fetchFavorites` — same pattern as `ActivitiesPage`. The favorited template activity ID is matched against the `activityId` field on the schedule entries.
- Sport options are derived dynamically from the activities list, so adding new sports to the backend doesn't require a frontend change.

### DIFF
--- a/client/src/pages/SchedulePage.jsx
+++ b/client/src/pages/SchedulePage.jsx
@@ -1,8 +1,10 @@
 import React, { useEffect, useMemo, useState } from 'react'
 import Button from '../components/ui/Button.jsx'
+import ScheduleFilters from '../components/ScheduleFilters.jsx'
 
 import { useNavigate } from 'react-router-dom'
 import { useAuth } from '../context/AuthContext.jsx'
+import { fetchFavorites } from '../services/FavoritesService.js'
 
 // ─────────────────────────────────────────────────────────────
 // SchedulePage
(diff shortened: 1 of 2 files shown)


isaqelle (approved):
This is so solid, love it, nice job!
One small thing I noticed: if fetching favorites fails, enabling the “Favorites only” filter would just show an empty schedule without any indication something went wrong. Maybe add a small error/toast state when favorites fail to load.

client/src/pages/SchedulePage.jsx
```diff
@@ -13,180 +15,292 @@ import { useAuth } from '../context/AuthContext.jsx'
...
 
   // ── Current Date ──────────────────────────────────────────
-  
-    const today = new Date()
-    // Test for JUNE
-    //const today = new Date('2026-06-15')
-    
-// ── Fetch Current Schedule ───────────────────────────────
-useEffect(() => {
+  const today = new Date()
```
jasminerezai, line 135: `today` is already declared at the top of the component, this one shadows it. Remove this line.

client/src/pages/SchedulePage.jsx
```diff
@@ -13,180 +15,292 @@ import { useAuth } from '../context/AuthContext.jsx'
...
+        setFavoriteIds([])
+        return
+      }
+
+      try {
+        const result = await fetchFavorites(token)
+        const ids = result.data.map(activity => activity.id)
+        setFavoriteIds(ids)
+      } catch (error) {
+        console.error('Failed to fetch favorites:', error)
```
jasminerezai, line 256: If fetching favorites fails, enabling "Favorites only" silently shows an empty schedule. Add a `favoritesError` state, set it to true in the catch block, and show a small error message when the filter is active but favorites failed to load.

jasminerezai (changes requested):
Great job Lawrence, just two small comments

client/src/pages/SchedulePage.jsx
```diff
@@ -17,180 +19,295 @@ import { API_BASE_URL } from '../services/apiConfig.js'
...
-    try {
+    async function fetchSchedule() {
 
+      try {
       const response = await fetch(
         `${API_BASE_URL}/api/schedules/current`
       )
 
-      const {data} = await response.json()
+        const response = await fetch('/api/schedules/current')
```
jasminerezai: I found this issue in Vercel for why it fails. Remove this line and the deployment will get built again

jasminerezai (approved):
Good job

PR #70, opened 2026-05-19
feat: added attendee list view for leader

### LINKED ISSUE #26: Leader: attendee list view
On the activity detail page, show leaders a list of registered participants.
Backend: GET /activities/:id/participants (leader only).

jasminerezai:
Closes #26

Added a leader attendee list view on the activity detail page. Leaders, board members, and admins can see a list of registered participants grouped by schedule for each activity.

Backend
GET /api/activities/:activityId - new endpoint to fetch a single activity by id (open to authenticated users)
GET /api/activities/:activityId/participants - new endpoint returning all schedules for an activity with their registered participants, restricted to LEADER and above. Leaders must own the activity, BOARD_MEMBER/ADMIN bypass the ownership check.

Frontend
New ActivityDetailPage.jsx — shows activity details (name, description, location, capacity, time slots) and an attendee list section for leaders
ActivitiesPage.jsx - activity names now link to the detail page
SchedulePage.jsx - activity card titles now link to the detail page
App.jsx - replaced placeholder with ActivityDetailPage

How to test
Log in as a leader (e.g. user2@example.com / user2pass)
Navigate to an activity they lead
Verify activity details are shown
Verify attendee list is visible grouped by schedule
Log in as a member and verify attendee list is not shown

### DIFF
--- a/client/src/pages/ActivityDetailPage.jsx
+++ b/client/src/pages/ActivityDetailPage.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react'
+import React, { useState, useEffect, useMemo } from 'react'
 import { useParams, Link, useNavigate } from 'react-router-dom'
 import { useAuth } from '../context/AuthContext.jsx'
 import Card from '../components/ui/Card.jsx'
@@ -17,17 +17,16 @@ export default function ActivityDetailPage() {
   useEffect(() => {
     async function loadActivity() {
       try {
-        const response = await fetch('/api/activities')
+        const headers = {}
+        if (token) {
+          headers['Authorization'] = `Bearer ${token}`
+        }
+        const response = await fetch(`/api/activities/${id}`, { headers })
         if (!response.ok) {
           throw new Error('Failed to fetch activity details')
         }
         const result = await response.json()
-        const found = result.data.find(act => act.id === id)
-        if (!found) {
-          setError('Activity not found')
-        } else {
-          setActivity(found)
-        }
+        setActivity(result.data)
       } catch (err) {
         setError(err.message)
       } finally {
@@ -35,14 +34,17 @@ export default function ActivityDetailPage() {
       }
     }
     loadActivity()
-  }, [id])
+  }, [id, token])
 
   // ── Determine if user is a leader/admin for this activity ────
-  const isLeader = isAuthenticated && (
-    user.role === 'ADMIN' ||
-    user.role === 'BOARD_MEMBER' ||
-    (user.role === 'LEADER' && activity?.leaders?.some(l => l.profileId === user.id))
-  )
+  const isLeader = useMemo(() => {
+    if (!isAuthenticated || !activity) return false
+    return (
+      user.role === 'ADMIN' ||
+      user.role === 'BOARD_MEMBER' ||
+      (user.role === 'LEADER' && activity.leaders?.some(l => l.profileId === user.id))
+    )
+  }, [isAuthenticated, user, activity])
 
   // ── Fetch Attendee List (Leaders/Admins only) ────────────────
   useEffect(() => {
@@ -106,7 +108,7 @@ export default function ActivityDetailPage() {
             </p>
           </div>
           {/* Action buttons if logged in */}
-          {isAuthenticated && (user.role === 'ADMIN' || user.role === 'BOARD_MEMBER' || (user.role === 'LEADER' && activity.leaders?.some(l => l.profileId === user.id))) && (
+          {isAuthenticated && (user.role === 'ADMIN' || user.role === 'BOARD_MEMBER') && (
             <Button as={Link} to={`/activities/${activity.id}/edit`} variant="outline" size="sm">
               Edit Activity
             </Button>
--- a/client/src/pages/ActivityDetailPage.jsx
+++ b/client/src/pages/ActivityDetailPage.jsx
@@ -11,6 +11,7 @@ export default function ActivityDetailPage() {
   const { isAuthenticated, user, token } = useAuth()
   const [activity, setActivity] = useState(null)
   const [participantData, setParticipantData] = useState(null)
+  const [participantError, setParticipantError] = useState(null)
   const [loading, setLoading] = useState(true)
   const [error, setError] = useState(null)
 
@@ -54,7 +55,7 @@ export default function ActivityDetailPage() {
                 setParticipantData(partResult.data)
               }
             } catch (partErr) {
-              // Silently ignore participant loading errors so the main details still render
+              setParticipantError('Could not load attendee list. Try refreshing.')
             }
           }
         }
@@ -95,28 +96,6 @@ export default function ActivityDetailPage() {
           <h2 style={{ color: 'var(--color-danger)', marginBottom: '12px' }}>Error</h2>
           <p style={{ marginBottom: '24px' }}>{error || 'Activity not found.'}</p>
           <Button onClick={() => navigate('/activities')}>Back to Activities</Button>
-          {activity.defaultStatus === 'CANCELLED' && (
-            <div style={{
-              background: 'var(--color-danger-light)',
-              borderLeft: '6px solid var(--color-danger)',
-              color: 'var(--color-danger)',
-              padding: 'var(--space-4) var(--space-6)',
-              borderRadius: 'var(--radius-md)',
-              marginBottom: 'var(--space-6)',
-              display: 'flex',
-              alignItems: 'center',
-              gap: 'var(--space-3)',
-              fontWeight: 700,
-            }}>
-              <span>⚠️</span>
-              <div>
-                <p style={{ margin: 0, fontWeight: 700 }}>This activity has been cancelled.</p>
-                <p style={{ margin: '2px 0 0', fontWeight: 400, fontSize: 'var(--text-sm)', opacity: 0.9 }}>
-                  No active schedules or future sessions will run until the activity is reactivated.
-                </p>
-              </div>
-            </div>
-          )}
         </Card>
       </div>
     )
@@ -128,6 +107,29 @@ export default function ActivityDetailPage() {
       <Link to="/activities" style={{ color: 'var(--color-primary-dark)', textDecoration: 'none', fontWeight: 600, display: 'inline-flex', alignItems: 'center', marginBottom: 'var(--space-4)' }}>
         ← Back to Activities
       </Link>
+
+      {activity.defaultStatus === 'CANCELLED' && (
+        <div style={{
+          background: 'var(--color-danger-light)',
+          borderLeft: '6px solid var(--color-danger)',
+          color: 'var(--color-danger)',
+          padding: 'var(--space-4) var(--space-6)',
+          borderRadius: 'var(--radius-md)',
+          marginBottom: 'var(--space-6)',
+          display: 'flex',
+          alignItems: 'center',
+          gap: 'var(--space-3)',
+          fontWeight: 700,
+        }}>
+          <span>⚠️</span>
+          <div>
+            <p style={{ margin: 0, fontWeight: 700 }}>This activity has been cancelled.</p>
+            <p style={{ margin: '2px 0 0', fontWeight: 400, fontSize: 'var(--text-sm)', opacity: 0.9 }}>
+              No active schedules or future sessions will run until the activity is reactivated.
+            </p>
+          </div>
+        </div>
+      )}
 
       <Card padding="md" shadow="md" style={{ marginBottom: 'var(--space-6)' }}>
         <div style={{ display: 'flex', justifyContent: 'space-between', alignItems: 'flex-start', flexWrap: 'wrap', gap: '16px' }}>
@@ -186,6 +188,12 @@ export default function ActivityDetailPage() {
           <p style={{ color: 'var(--color-text-muted)', fontSize: '0.9rem', marginBottom: '20px' }}>
             Registered participants for all schedules of this activity. Only visible to leaders.
           </p>
+
+          {participantError && (
+            <p style={{ color: 'var(--color-danger)', fontSize: '0.9rem', marginBottom: '16px' }}>
+              {participantError}
+            </p>
+          )}
 
           {!participantData || participantData.schedules?.length === 0 ? (
             <p style={{ fontStyle: 'italic', color: 'var(--color-text-muted)' }}>No schedules found for this activity.</p>
--- a/client/src/pages/ActivityDetailPage.jsx
+++ b/client/src/pages/ActivityDetailPage.jsx
@@ -13,28 +13,59 @@ export default function ActivityDetailPage() {
   const [loading, setLoading] = useState(true)
   const [error, setError] = useState(null)
 
-  // ── Fetch Activity Details ───────────────────────────────────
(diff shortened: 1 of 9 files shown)


server/src/controllers/activities.controller.ts
```diff
@@ -257,3 +257,80 @@ export const unregisterParticipation = asyncHandler(async (
 
   res.status(200).json({ status: 'success', data: { participantCount } })
 })
+
+export const getActivityParticipants = asyncHandler(async (
+  req: Request<{ activityId: string }>,
+  res: Response<ApiResponse<any>>
```
JuriRappold:
```suggestion
  res: Response<ApiResponse<ScheduleWPartsDto>>
```
create a type that adheres to your currently sent back object:
```typescript
type ParticipantsDto = Pick<Profile, 'id' | 'profileName' | 'email'>

type ScheduleWPartsDto = {
    scheduleId: string,
    startAt: string, //not sure rn, if its string or date for the starting/ending times
    endAt: string,
    status: ActivityStatus,
    participations: ParticipantsDto
}
```

server/src/controllers/activities.controller.ts
```diff
@@ -257,3 +257,80 @@ export const unregisterParticipation = asyncHandler(async (
...
+      activityId: activity.id,
+      activityName: activity.name,
+      schedules: formattedSchedules
+    }
+  })
+})
+
+export const getActivityById = asyncHandler(async (
+  req: Request<{ activityId: string }>,
+  res: Response<ApiResponse<any>>
```
JuriRappold:
```suggestion
  res: Response<ApiResponse<ActivityDto>>
```

JuriRappold (changes requested):
Noice PR! Just how your response object looks like, you still have the `any` response type in it ( `res: Response< ApiResponse< any > >`).
But otherwise nice job!

larrymijo (approved):

server/src/controllers/activities.controller.ts
```diff
@@ -257,3 +257,80 @@ export const unregisterParticipation = asyncHandler(async (
...
+  // 3. Format response
+  const formattedSchedules = schedules.map(s => ({
+    scheduleId: s.id,
+    startAt: s.startAt,
+    endAt: s.endAt,
+    status: s.status,
+    participants: s.participations.map(p => ({
+      id: p.profile.id,
+      profileName: p.profile.profileName || 'No Name',
+      email: p.profile.email
```
gunnarwrld, line 299: The participants response includes:

  `email: p.profile.email`

A LEADER assigned to activity A can call `GET /activities/B/participants` if
they somehow get activity B's ID, but `assertActivityAccess` blocks that.

However, even for their own activity, exposing raw email addresses in an API
response is a privacy concern. The frontend only renders `profileName` in the
participant list anyway (email is shown in parentheses as secondary info).

If email is intentionally shown to leaders, add a comment justifying it.
If it's not needed, remove it from the select and the formatted response.

server/src/routes/activities.routes.ts
```diff
@@ -31,9 +33,20 @@ router.patch(
...
+  authMiddleware,
+  restrictToMinRole(ProfileRole.LEADER),
+  getActivityParticipants
+);
+
 // GET /activities
 router.get('', getActivities);
 
+// GET /activities/:activityId --- anyone logged in --> get single activity details
+router.get('/:activityId', authMiddleware, getActivityById);
```
gunnarwrld: This means unauthenticated users (not logged in) will get a 401 when visiting an activity detail page directly (e.g. via a shared link). The comment says "anyone logged in" but I suggest it should be public.

If we make the login required: fix the frontend to always send the token and handle the 401 redirect.

server/src/controllers/activities.controller.ts
gunnarwrld, line 1: Both `getActivityParticipants` and `getActivityById` use `Response<ApiResponse<any>>`.
Define proper return types. For `getActivityById`, `ActivityDto` already exists.
For `getActivityParticipants`, define a dedicated DTO:
```
  type ParticipantScheduleDto = {
    scheduleId: string
    startAt: Date
    endAt: Date | null
    status: ActivityStatus
    participants: { id: string; profileName: string | null; email: string }[]
  }

  type ActivityParticipantsDto = {
    activityId: string
    activityName: string
    schedules: ParticipantScheduleDto[]
  }
```
Add it to `types/index.ts` and use it on the response type.

server/src/db/readQueries.ts
```diff
@@ -142,9 +142,13 @@ export class READ {
...
+    static async activityById(activityId: string) {
         const activity = await prisma.activityTemplate.findUnique({
-            where: { id: activityId }
+            where: { id: activityId },
+            include: {
+                timeSlots: true,
+                leaders: true
+            }
         });
         return activity;
```
gunnarwrld: The explicit return type `Promise<ActivityTemplate | null>` was removed and replaced with implicit inference. The return type is now whatever Prisma infers from the include, which is correct but loses the contract at the
call site. Add an explicit return type using the inferred Prisma type or`ActivityDto` if it matches.

gunnarwrld: The explicit return type `Promise<ActivityTemplate | null>` was removed and replaced with implicit inference. The return type is now whatever Prisma infers from the include, which is correct but loses the contract at the
call site. Add an explicit return type using the inferred Prisma type or`ActivityDto` if it matches.

gunnarwrld (changes requested):
The auth gate on `GET /:activityId` is the most impactful, it will break the page for unauthenticated users hitting a direct link. The `ApiResponse<any>` and email exposure need addressing before merge.

server/src/controllers/activities.controller.ts
jasminerezai: @JuriRappold Thanks  Juri for the suggestion. Since the controller formats the response wrapped in a top-level object containing the activity metadata (i.e., activityId and activityName), and defines participants as an array (rather than a singular object named participations), we followed @gunnarwrld's suggestion to use ActivityParticipantsDto and ParticipantScheduleDto which matches the exact response shape perfectly

server/src/controllers/activities.controller.ts
jasminerezai: Nice catch! I've updated the return type of getActivityById to Response<ApiResponse<ActivityDto>>

server/src/controllers/activities.controller.ts
jasminerezai: Exposing raw emails to leaders is intentional since they need them to directly contact or coordinate with the participants registered for their schedules (especially for coordinates, cancellation alerts, or changes). I've added a code comment in the controller to clearly justify this exposure.

server/src/routes/activities.routes.ts
jasminerezai: Completely agree, the page should indeed be public! I've removed authMiddleware from GET /activities/:activityId so direct links don't break for unauthenticated users.

server/src/db/readQueries.ts
jasminerezai: Good call on maintaining the contract at the call site. I've added explicit return types using ActivityDto to both activityById and allActivities

client/src/pages/ActivityDetailPage.jsx
```diff
@@ -0,0 +1,224 @@
...
+        const response = await fetch(`/api/activities/${id}/participants`, {
+          headers: {
+            Authorization: `Bearer ${token}`,
+          },
+        })
+        if (response.ok) {
+          const result = await response.json()
+          setParticipantData(result.data)
+        } else {
+          console.error('Failed to load participants', response.statusText)
```
gunnarwrld: Two `console.error` calls left in the fetch error paths:
```
  console.error('Failed to load participants', response.statusText)  // line 64
  console.error('Error fetching participants:', err)                  // line 67
```
These were flagged in the previous round for `activities.controller.ts` and
fixed there, but the same pattern now exists in the new frontend file.
Remove both, the error state is already set via `setError` in the catch block.

client/src/pages/ActivityDetailPage.jsx
```diff
@@ -0,0 +1,224 @@
...
+  const [participantData, setParticipantData] = useState(null)
+  const [loading, setLoading] = useState(true)
+  const [error, setError] = useState(null)
+
+  // ── Fetch Activity Details ───────────────────────────────────
+  useEffect(() => {
+    async function loadActivity() {
+      try {
+        const headers = {}
+        if (token) {
```
gunnarwrld, line 26: The route `GET /activities/:id` in `App.jsx` is no longer wrapped in `ProtectedRoute`
so the endpoint must be public. But the backend route still has `authMiddleware`
on it (if you could change it in this same PR would be awesome!).

client/src/pages/ActivityDetailPage.jsx
```diff
@@ -0,0 +1,224 @@
...
+    loadActivity()
+  }, [id, token])
+
+  // ── Determine if user is a leader/admin for this activity ────
+  const isLeader = useMemo(() => {
+    if (!isAuthenticated || !activity) return false
+    return (
+      user.role === 'ADMIN' ||
+      user.role === 'BOARD_MEMBER' ||
+      (user.role === 'LEADER' && activity.leaders?.some(l => l.profileId === user.id))
```
gunnarwrld, line 80: This works correctly with the current `ActivityDto.leaders` shape. However this
is the exact inconsistency flagged in the normalize leaders issue, if `ActivityDto.leaders` is ever normalized to `Pick<Profile, 'id' | 'profileName'>[]` (issue opened today), this line will silently break because the field will be `l.id` not `l.profileId`.

Add a comment here linking to that issue so whoever does the normalization (i could do it) knows to update this check:
```
  // NOTE: ActivityDto.leaders uses { profileId, activityId } shape.
  // If normalized to { id, profileName } (see issue #XX), change to l.id === user.id
```

client/src/pages/ActivityDetailPage.jsx
gunnarwrld, line 1: `loadActivity` fires on `[id, token]`.
`loadParticipants` fires on `[id, isLeader, token]`.

`isLeader` depends on activity being loaded (it memos from `activity.leaders`).
So the entire thing will be:
  1. loadActivity` fires, then activity set and finally `isLeader` recomputes
  2. `loadParticipants` fires (isLeader now true for a leader)

This works but only because React batches state updates. If activity fetch is
slow, `isLeader` is false on the first render and `loadParticipants` exits early
(`if (!isLeader) return`). 

It will re-run when `isLeader updates`, so the final result is correct. But it means a leader always triggers two
render cycles for the participants fetch.

Not a blocker but worth noting as a potential source of flicker or double fetch. Consider consolidating into a single `useEffect` or fetching participants inside `loadActivity` after the activity is confirmed.

server/src/controllers/activities.controller.ts
```diff
@@ -257,3 +257,82 @@ export const unregisterParticipation = asyncHandler(async (
...
+) => {
+  const { activityId } = req.params
+  const { id: requesterId, role } = req.user!
+
+  if (!isUUID(activityId)) {
+    throw ApiError.badRequest('Invalid activityId format')
+  }
+
+  // 1. Verify existence + leadership/admin access
+  const activity = await assertActivityAccess(activityId, requesterId, role)
```
gunnarwrld, line 283: `assertActivityAccess` fetches the `activityTemplate` to verify ownership. Then immediately after, the handler fires a separate `prisma.schedule.findMany` for the same `activityId`. That's 2 DB round trips
where 1 would do if the schedule fetch was moved into a query layer method (similar to how `getActivityById` was refactored into `READ.activityById`).

Not a blocker for this PR but inconsistent with the pattern of keeping Prisma calls out of controllers.

gunnarwrld (commented):
Good iteration! All of the previous round's issues are addressed: the DTOs are properly typed, the email exposure is justified with a clear comment, and the routing is cleaner. Just a couple of things that i suggest implementing before merging and also one thing to fix regarding the deletion of the `ProtectedRoute`.

client/src/pages/ActivityDetailPage.jsx
jasminerezai: it is actually already fully public on the backend (registered in `activities.routes.ts` without any `authMiddleware` or role checks), so it is completely ready for unauthenticated access!

jasminerezai:
Thanks for the excellent and detailed review, @gunnarwrld! 

I've resolved these points in the latest commit:
- **Hook Consolidation & Cleanup:** Consolidated the two frontend `useEffect` hooks in `ActivityDetailPage.jsx` into one unified effect. This fetches both the activity details and the participant list (if authorized) in one clean cycle before dismissing the loading state, entirely eliminating any rendering flicker or double fetch cycle. Also removed the redundant `console.error` calls.
- **Normalization Warning:** Added a clear warning comment to the `isLeader` memo to ensure whoever normalizes `ActivityDto.leaders` is prompted to update the checking property from `l.profileId` to `l.id`.
- **Backend Prisma Query Extraction:** Extracted the schedules query out of the `getActivityParticipants` controller into `READ.activityParticipants` in our DB queries layer, keeping all controller methods thin and consistent with our codebase pattern.

And already left a comment for this but here again. 
- **Public Endpoint Verification:** Regarding the `GET /activities/:id` route, it is actually already fully public on the backend (registered in `activities.routes.ts` without any `authMiddleware` or role checks), so it is completely ready for unauthenticated access!

JuriRappold (approved):
Looks good to me!
For Dtos in general i was thinking of using the prisma types, kinda how i did it with the Schedules. What do you guys think about that. I do like the explicit Dto too though, easier to read.

jasminerezai:
@JuriRappold 
Agreed the explicit DTOs are easier to read and reason about - keeps the API contract clear and decoupled from Prisma internals. Happy to keep it as is. The PR just needs the merge conflicts resolved and it's ready to merge.

isaqelle (approved):

client/src/pages/ActivityDetailPage.jsx
```diff
@@ -0,0 +1,233 @@
...
+              const partResponse = await fetch(`/api/activities/${id}/participants`, {
+                headers: {
+                  Authorization: `Bearer ${token}`,
+                },
+              })
+              if (partResponse.ok) {
+                const partResult = await partResponse.json()
+                setParticipantData(partResult.data)
+              }
+            } catch (partErr) {
```
gunnarwrld, line 57: This means if the participants request fails (expired token, 403, network error),
the leader sees an empty attendee list with no indication of why. 
From a frontend perspective this is worse than a inline error, the leader might think
there are genuinely zero participants.

Maybe we can set a dedicated participants error state and rendering a small
inline message in the attendee section:
```
  } catch (partErr) {
    setParticipantError('Could not load attendee list. Try refreshing.')
  }
```
Not a blocker but worth addressing before this goes to production.

gunnarwrld (approved):
Everything correct on this PR, just one recommendation and also i was thinking it would be good to create and assign an issue related to track for the responsive design on the frontend.

jasminerezai:
@gunnarwrld Already created an issue for that and assigned it to Isabelle!

client/src/pages/ActivityDetailPage.jsx
jasminerezai: fixed before merged

PR #72, opened 2026-05-19
Fix: API calls failing in production (frontend couldn't reach backend)

### LINKED ISSUE #71: API calls failing in production (frontend couldn't reach backend)

larrymijo:
### Fix: API calls failing in production (frontend couldn't reach backend)

Centralized the backend base URL so the frontend talks to the deployed backend in production. Every `fetch('/api/...')` call now goes through `${API_BASE_URL}/api/...` where `API_BASE_URL` reads from `VITE_API_URL` at build time.

### What happened
All our fetch calls were relative (`/api/activities`, `/api/auth/login`, etc.). That works locally because of the vite proxy in `vite.config.js`, but the proxy only exists in `vite dev` — it doesn't get bundled. So on Vercel the requests were going to `vercel.app/api/...`, which doesn't exist, and nothing was loading.

### Changes
- New file: `client/src/services/apiConfig.js` — single source of truth for the base URL. Falls back to `''` when `VITE_API_URL` isn't set so local dev keeps working without any setup.
- Updated `AuthService.js`, `FavoritesService.js`, `SchedulePage.jsx`, `ActivitiesPage.jsx`, `ActivityFormPage.jsx` to use it.
- Updated the usage example in `AuthContext.jsx` comments so we don't copy the broken pattern next time.
- Added `client/.env.example` so it's clear what env var to set.

### Setup needed before this works in prod
`VITE_API_URL` has to be set in the Vercel project's env vars (pointing at the deployed backend), then redeployed. Vite inlines env vars at build time so a redeploy is required for changes to take effect.

Local dev needs no setup — just don't create a `client/.env` and the vite proxy handles everything.

### How to test
- **Locally**: `cd client && npm run dev`, backend running on `:3001`, log in / browse activities / schedule — should all work as before.
- **Prod**: after Vercel env var is set + redeploy, open the deployed URL and check the network tab — `/api/*` requests should hit the deployed backend, not the vercel.app domain.

### Notes for reviewers
- I kept the per-file `BASE` constants in `AuthService` and `FavoritesService` since they group related routes nicely — they just now derive from `API_BASE_URL`.
- One thing I noticed but didn't touch: the role check in `ActivityFormPage` runs *after* the `useEffect` that loads the activity, so unauthorized users still fire one fetch on mount. Not a security issue (backend rejects it), but might be worth a quick follow-up.

### DIFF
--- a/client/src/services/apiConfig.js
+++ b/client/src/services/apiConfig.js
@@ -1,0 +1,26 @@
+// Single source of truth for the backend base URL.
+//
+// Why this file exists:
+//   In dev, vite.config.js proxies '/api/*' to http://localhost:3001
+//   so relative URLs like fetch('/api/activities') just work.
+//   In production (Vercel) there is NO proxy — the frontend and
+//   backend are on different origins, so we must prefix every
+//   request with the deployed backend URL.
+//
+// Usage:
+//   import { API_BASE_URL } from '../services/apiConfig.js'
+//   fetch(`${API_BASE_URL}/api/activities`)
+//
+// How VITE_API_URL is set:
+//   - Local dev:   leave it unset → falls back to '' → vite proxy handles it
+//   - Production:  set VITE_API_URL in Vercel project settings to the
+//                  deployed backend origin, e.g. https://hkif-api.onrender.com
+//                  (NO trailing slash, NO '/api' suffix)
+//
+// Note: Vite inlines env vars at BUILD time, not runtime.
+// If you change VITE_API_URL on Vercel you must redeploy.
+// ─────────────────────────────────────────────────────────────
+
+export const API_BASE_URL = import.meta.env.VITE_API_URL ?? ''
+// '' (empty string) keeps relative paths working in dev, which means
+// the vite proxy in vite.config.js still kicks in and forwards to localhost:3001.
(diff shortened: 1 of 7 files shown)


client/src/services/apiConfig.js
```diff
@@ -0,0 +1,26 @@
...
+//   - Local dev:   leave it unset → falls back to '' → vite proxy handles it
+//   - Production:  set VITE_API_URL in Vercel project settings to the
+//                  deployed backend origin, e.g. https://hkif-api.onrender.com
+//                  (NO trailing slash, NO '/api' suffix)
+//
+// Note: Vite inlines env vars at BUILD time, not runtime.
+// If you change VITE_API_URL on Vercel you must redeploy.
+// ─────────────────────────────────────────────────────────────
+
+export const API_BASE_URL = import.meta.env.VITE_API_URL ?? ''
```
jasminerezai, line 24: Great, exactly what I meant

jasminerezai (approved):
Good, looks correct! Exactly the way I want it to be. Thanks a lot

gunnarwrld:
That is perfect and yes i have mention this multiple times: The role check in `ActivityFormPage` runs after the `useEffect `that loads the activity. We have to solve that, ill do it myself

PR #73, opened 2026-05-19
feat: add cancellation badge on activity cards and detail page

### LINKED ISSUE #28: Cancellation notice on activity cards
If an activity is cancelled, show a visual indicator (badge or banner) on the card and detail page.
Backend: add a `cancelled` boolean field to Activity model if not already there.

jasminerezai:
Closes #28

Added cancellation indicators across the app so users can clearly see when an activity has been cancelled.
Changes

ActivitiesPage.jsx - added a red "Cancelled" badge next to the activity name when defaultStatus === 'CANCELLED'
ActivityDetailPage.jsx - added a prominent warning banner at the top of the detail page when the activity is cancelled, plus a status badge next to the activity title

No backend changes needed - defaultStatus with CANCELLED value already exists in the database.
How to test

Set an activity's defaultStatus to CANCELLED in the database
Go to /activities and verify the red Cancelled badge appears on the card
Click into the activity and verify the warning banner appears at the top of the detail page
Set the status back to ACTIVE and verify both indicators disappear

### DIFF
--- a/client/src/pages/ActivitiesPage.jsx
+++ b/client/src/pages/ActivitiesPage.jsx
@@ -4,7 +4,7 @@ import { useAuth } from '../context/AuthContext.jsx'
 import { fetchFavorites, addFavorite, removeFavorite, } from '../services/FavoritesService.js'
 import Card from '../components/ui/Card.jsx'
 import Button from '../components/ui/Button.jsx'
-import Badge from '../components/ui/Badge.jsx'
+import Badge, { STATUS_VARIANT } from '../components/ui/Badge.jsx'
 
 export default function ActivitiesPage() {
   // ── API State ─────────────────────────────────────────────
@@ -184,7 +184,7 @@ export default function ActivitiesPage() {
               <h2 style={{ marginBottom: '8px', display: 'flex', alignItems: 'center', gap: 'var(--space-2)', flexWrap: 'wrap' }}>
                 {activity.name}
                 {activity.defaultStatus === 'CANCELLED' && (
-                  <Badge variant="danger">Cancelled</Badge>
+                  <Badge variant={STATUS_VARIANT[activity.defaultStatus]}>Cancelled</Badge>
                 )}
               </h2>
 
--- a/client/src/pages/ActivitiesPage.jsx
+++ b/client/src/pages/ActivitiesPage.jsx
@@ -4,6 +4,7 @@ import { useAuth } from '../context/AuthContext.jsx'
 import { fetchFavorites, addFavorite, removeFavorite, } from '../services/FavoritesService.js'
 import Card from '../components/ui/Card.jsx'
 import Button from '../components/ui/Button.jsx'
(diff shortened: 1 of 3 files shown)


client/src/pages/ActivityDetailPage.jsx
```diff
@@ -0,0 +1,215 @@
...
+              }}>
+                {activity.name}
+              </h1>
+              <p style={{ color: 'var(--color-text-muted)', fontSize: 'var(--text-sm)', marginTop: 'var(--space-1)' }}>
+                Activity Details
+              </p>
+            </div>
+            
+            {/* Status Badge */}
+            <Badge variant={isCancelled ? 'danger' : 'success'}>
```
larrymijo: This treats defaultStatus as a boolean (cancelled / not-cancelled), but the ActivityStatus enum in schema.prisma has four values: ACTIVE, INACTIVE, CANCELLED, DELAYED. The form in ActivityFormPage.jsx (line 7) already lets admins set all four. With this code, an INACTIVE or DELAYED activity will show a green "Active" badge, which is misleading.
Could we use the STATUS_VARIANT map already exported from Badge.jsx and just render the actual status? Something like:
import Badge, { STATUS_VARIANT } from '../components/ui/Badge.jsx'
...
<Badge variant={STATUS_VARIANT[activity.defaultStatus]}>
  {activity.defaultStatus}
</Badge>
That way all four states render correctly with no extra work, and we stay consistent with the convention documented in Badge.jsx.

client/src/pages/ActivitiesPage.jsx
```diff
@@ -180,8 +181,11 @@ export default function ActivitiesPage() {
                   ? '♥'
                   : '♡'}
               </button>
-              <h2 style={{ marginBottom: '8px' }}>
+              <h2 style={{ marginBottom: '8px', display: 'flex', alignItems: 'center', gap: 'var(--space-2)', flexWrap: 'wrap' }}>
                 {activity.name}
+                {activity.defaultStatus === 'CANCELLED' && (
```
larrymijo, line 187: Same as in ActivityDetailPage, we could use the STATUS_VARIANT helper from Badge.jsx here so we don't hardcode the variant string. The pattern is even documented in Badge.jsx lines 50-53. Keeps the mapping in one place if we ever change the colour scheme. Not blocking, just a consistency nit.

larrymijo (commented):
nice work, the banner looks clean and the flow makes sense 👍 left two small things — the main one is that the detail page status badge only handles cancelled/active, but the enum has 4 states (INACTIVE and DELAYED would currently show as green "Active"). also we already have a STATUS_VARIANT map in Badge.jsx that does the mapping for us, would be nice to use it instead of hardcoding "danger"

isaqelle:
Nice job! I agree with Lawrence on the changes

larrymijo (approved):
approved the changes!

isaqelle (approved):

PR #75, opened 2026-05-19
feat: add role-based route protection via requiredRoles prop

### LINKED ISSUE #74: feature/protected-route-roles

larrymijo:
### What
Extended `ProtectedRoute` with an optional `requiredRoles` prop so role checks live in the route config instead of being scattered inside individual pages. Used it on the activity create/edit routes and removed the now-redundant inline check from `ActivityFormPage`.

### Why
In `ActivityFormPage` the role check rendered *after* the `useEffect` that fetches the activity, so a logged-in MEMBER hitting `/activities/5/edit` would still trigger a `GET /api/activities/5` before being shown the "no permission" message. Wasteful, and the dead-end error page was bad UX.

Not a security issue — the backend enforces roles via middleware regardless of what the frontend does — but worth fixing before we add more leader/admin-only pages and end up copy-pasting the check around.

### Changes
- **New** `client/src/constants/roles.js` — `ROLES` map and `MANAGER_ROLES` array so role names live in one place. Stays in sync with the `ProfileRole` enum in `prisma/schema.prisma`.
- **`App.jsx`** — `ProtectedRoute` now accepts `requiredRoles`. Logged-in users with the wrong role get redirected to `/` (home) instead of seeing an error page. When `requiredRoles` is omitted the component behaves exactly as before (auth-only), so existing usage on `/profile` is unchanged.
- **`App.jsx`** — `/activities/new` and `/activities/:id/edit` now use `requiredRoles={MANAGER_ROLES}`.
- **`ActivityFormPage.jsx`** — dropped the inline `canManageActivities` check + early return. Also removed `user` from the `useAuth` destructure since it's no longer referenced.

### Notes for reviewers
- Redirect target for unauthorized users is `/` for now. If we want a dedicated `/forbidden` page later it's a one-line change.
- The constants file is small on purpose — added it now to avoid repeating `['LEADER', 'BOARD_MEMBER', 'ADMIN']` across routes. Open to renaming `MANAGER_ROLES` if anyone has a clearer name.
- Backend role middleware is unchanged — this PR only affects the client-side guard.

### DIFF
--- a/client/src/App.jsx
+++ b/client/src/App.jsx
@@ -7,8 +7,8 @@ import LoginPage    from './pages/LoginPage.jsx'
 import RegisterPage from './pages/RegisterPage.jsx'
 import ActivityFormPage from './pages/ActivityFormPage.jsx'
 import SchedulePage from './pages/SchedulePage.jsx'
-import { MANAGER_ROLES } from './constants/roles.js'
- 
+import { MANAGER_ROLES, EDITOR_ROLES } from './constants/roles.js'
+
 // ── ProtectedRoute ────────────────────────────────────────────
 // Wraps any route that requires login, and optionally a specific role.
 //
@@ -19,16 +19,18 @@ import { MANAGER_ROLES } from './constants/roles.js'
 //                    Leave undefined for "logged-in only, any role".
 //
 // Redirects:
-//   Not logged in        → /login (preserves blocked URL in location.state)
-//   Logged in, wrong role → / (home)
+//   Still rehydrating         → spinner (no redirect yet)
+//   Not logged in             → /login (preserves blocked URL)
+//   Logged in but no role     → /login (treat as broken session)
+//   Logged in, wrong role     → / (home)
 //
-// Note: backend role middleware is the real enforcement layer — this guard
-// just prevents the page from mounting client-side so we avoid pointless
-// fetches and dead-end "no permission" screens.
+// Note: backend role middleware is the real enforcement layer — this
+// guard just prevents the page from mounting client-side so we avoid
+// pointless fetches and dead-end "no permission" screens.
 function ProtectedRoute({ children, requiredRoles }) {
   const { isAuthenticated, loading, user } = useAuth()
   const location = useLocation()
- 
+
   if (loading) {
     // Still checking localStorage — don't redirect yet.
     // Without this check, a logged-in user would be briefly
@@ -45,32 +47,43 @@ function ProtectedRoute({ children, requiredRoles }) {
       </div>
     )
   }
- 
+
   if (!isAuthenticated) {
+    // Not logged in (or token was rejected during rehydration).
+    // Preserve the blocked URL so LoginPage can return them here.
     return (
       <Navigate
         to="/login"
         replace
-        // replace: don't push the blocked URL onto history —
-        // back button won't loop them back to the blocked page
         state={{ from: location }}
-        // location.pathname passed here so LoginPage can redirect
-        // the user back after they log in
       />
     )
   }
- 
-  // Role check — only runs when requiredRoles is provided.
-  // user.role comes from the JWT-backed /me lookup in AuthContext.
-  // If a user with the wrong role lands here (e.g. via a shared link),
-  // bounce them to home rather than showing an inline error page.
-  if (requiredRoles && !requiredRoles.includes(user?.role)) {
+
+  // Defensive: if a route requires a role but the user object somehow
+  // doesn't have one, treat the session as broken and force a fresh
+  // login. Better than silently bouncing the user to home with no
+  // explanation (which is what `requiredRoles.includes(undefined)`
+  // would produce).
+  if (requiredRoles && !user?.role) {
+    return (
+      <Navigate
+        to="/login"
+        replace
+        state={{ from: location }}
+      />
+    )
+  }
+
+  // Logged in, but with the wrong role for this route.
+  // Bounce to home rather than showing an inline error page.
+  if (requiredRoles && !requiredRoles.includes(user.role)) {
     return <Navigate to="/" replace />
   }
- 
+
   return children
 }
- 
+
 // ── Placeholder pages ─────────────────────────────────────────
 // Replace these one by one as you build the real pages.
 function Placeholder({ title }) {
@@ -98,7 +111,7 @@ function Placeholder({ title }) {
     </div>
   )
 }
- 
+
 // ── AppRoutes ─────────────────────────────────────────────────
 // Separated from App so it can use useAuth() —
 // hooks only work inside the Provider tree.
@@ -110,9 +123,12 @@ function AppRoutes() {
         <Routes>
           {/* Public routes — anyone can access */}
           <Route path="/" element={<SchedulePage />} />
-          <Route path="/login"          element={<LoginPage />} />
-          <Route path="/register"       element={<RegisterPage />} />
+          <Route path="/login"    element={<LoginPage />} />
+          <Route path="/register" element={<RegisterPage />} />
           <Route path="/activities" element={<ActivitiesPage />} />
+
+          {/* Create form: BOARD_MEMBER+ only — backend POST /activities
+              is gated to the same set, so LEADERs would just hit a 403. */}
           <Route
             path="/activities/new"
             element={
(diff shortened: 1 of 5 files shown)


client/src/App.jsx
```diff
@@ -94,10 +113,24 @@ function AppRoutes() {
           <Route path="/login"          element={<LoginPage />} />
           <Route path="/register"       element={<RegisterPage />} />
           <Route path="/activities" element={<ActivitiesPage />} />
```
gunnarwrld, line 128: `ActivitiesPage.jsx` no longer exports `ActivitiesPage`  it now exports `ActivityFormPage`.

client/src/App.jsx
```diff
@@ -34,24 +45,32 @@ function ProtectedRoute({ children }) {
...
       />
     )
   }
-
+ 
+  // Role check — only runs when requiredRoles is provided.
+  // user.role comes from the JWT-backed /me lookup in AuthContext.
+  // If a user with the wrong role lands here (e.g. via a shared link),
+  // bounce them to home rather than showing an inline error page.
+  if (requiredRoles && !requiredRoles.includes(user?.role)) {
```
gunnarwrld: This runs after the loading check, so if loading is still true the role check
is skipped. But if loading is false and user is null (e.g. token expired mid-
session), `user?.role` is undefined, `requiredRoles.includes(undefined)` is false,
and the user gets silently bounced to "/" with no indication of why, they
should be redirected to `/login` instead.

client/src/constants/roles.js
```diff
@@ -0,0 +1,27 @@
...
+  LEADER:       'LEADER',
+  BOARD_MEMBER: 'BOARD_MEMBER',
+  ADMIN:        'ADMIN',
+}
+
+// Roles allowed to create, edit, and manage activities.
+// Used by ProtectedRoute on the /activities/new and /activities/:id/edit routes.
+// Backend enforces this independently via role middleware — this is the
+// frontend gate that stops the page from mounting at all.
+export const MANAGER_ROLES = [
```
gunnarwrld, line 36: A LEADER can reach `/activities/new` and `/activities/:id/edit`. Is that
intentional? 
In the backend, `createSchedule` and `updateSchedule` gate LEADERs
via `assertActivityAccess` (own activities only). But the create activity
endpoint (`POST /activities`) is restricted to `BOARD_MEMBER+` on the backend.

So a LEADER can reach the create form on the frontend but will get a 403 from
the backend on submit. 

Consider either:
- Removing `ROLES.LEADER` from `MANAGER_ROLES` and creating a separate
  `EDITOR_ROLES = [ROLES.LEADER, ROLES.BOARD_MEMBER, ROLES.ADMIN]` for edit-only routes
- Or keeping it and handling the 403 gracefully in `ActivityFormPage` with a
  user-facing message

client/src/pages/ActivitiesPage.jsx
```diff
@@ -1,230 +1,325 @@
...
-export default function ActivitiesPage() {
-  // ── API State ─────────────────────────────────────────────
-  const [activities, setActivities] = useState([])
-  const [favoriteActivities, setFavoriteActivities] = useState([])
-  const navigate = useNavigate()
-  const { isAuthenticated, token } = useAuth()
-  const [loading, setLoading] = useState(true)
-  const [error, setError] = useState(null)
+const WEEKDAYS = ['MONDAY', 'TUESDAY', 'WEDNESDAY', 'THURSDAY', 'FRIDAY', 'SATURDAY', 'SUNDAY']
+const STATUSES = ['ACTIVE', 'INACTIVE', 'CANCELLED', 'DELAYED']
```
gunnarwrld: Same issue flagged in PR #66 and PR #67 on the backend. The frontend now has
its own copy of the same hardcoded enum. If a status is added to the Prisma
schema, this silently diverges.

The backend already validates status server-side, it should be derived from a shared constant or at minimum moved into `roles.js` (a separate constants file so there's one place to update it).

client/src/pages/ActivitiesPage.jsx
```diff
@@ -1,230 +1,325 @@
...
+  // Submit
+  async function handleSubmit(e) {
+    e.preventDefault()
+    setErrors({})
+    setServerError('')
 
-    if (!isAuthenticated) {
-      navigate('/login')
+    const errs = validate()
+    if (Object.keys(errs).length > 0) {
```
gunnarwrld: `handleSubmit` fires `POST` or `PUT` with the auth header but only catches generic
errors. If the token is expired, the backend returns 401, the form will just
show a generic error. Add a 401 check that redirects to `/login` and clears the
auth context, consistent with how other pages handle it.

gunnarwrld (changes requested):
The `ActivitiesPage` export being gone while `App.jsx` still references it is a runtime crash. The null user edge case on the role check redirects to the wrong place. LEADER in MANAGER_ROLES for the create route needs a decision, either fix the UX or document it explicitly.

larrymijo:
> The `ActivitiesPage` export being gone while `App.jsx` still references it is a runtime crash. The null user edge case on the role check redirects to the wrong place. LEADER in MANAGER_ROLES for the create route needs a decision, either fix the UX or document it explicitly.

## Made some changes
- **New `client/src/constants/roles.js`** — `ROLES` map plus two role
  groups that mirror the backend's permission model:
  - `MANAGER_ROLES = [BOARD_MEMBER, ADMIN]` for routes that create
    (backend `POST /activities` is gated to BOARD_MEMBER+).
  - `EDITOR_ROLES  = [LEADER, BOARD_MEMBER, ADMIN]` for routes that
    edit (backend `PUT /activities/:id` lets LEADERs update their own
    activities — own-only check stays on the backend).
- **New `client/src/constants/activityEnums.js`** — `ACTIVITY_STATUSES`
  and `WEEKDAYS` pulled out of the form so the schema and the dropdowns
  share one source on the frontend.
- **`App.jsx`** — `ProtectedRoute` now accepts `requiredRoles`. Redirect
  logic:
  - Not authenticated → `/login` (preserving the blocked URL).
  - Authenticated but `user.role` is missing → `/login` (defensive,
    treat the session as broken).
  - Authenticated, wrong role → `/` (home).
  - When `requiredRoles` is omitted the component behaves exactly as
    before (auth-only), so existing usage on `/profile` is unchanged.
- **`App.jsx`** — `/activities/new` uses `MANAGER_ROLES`,
  `/activities/:id/edit` uses `EDITOR_ROLES`.
- **`ActivityFormPage.jsx`** — dropped the inline role check + early
  return, removed `user` from `useAuth` since it's no longer referenced,
  pulled status/weekday lists from the new constants file, and added 401
  handling to both the load `useEffect` and `handleSubmit`. On 401 the
  form clears the auth context via `logout()` and bounces to `/login`
  with the current path preserved (consistent with how rehydration
  handles an invalid token).
- **`ActivitiesPage.jsx`** — restored from main; it had been accidentally
  overwritten with a duplicate of `ActivityFormPage` in the first push.

## Notes for review
- Redirect target for "logged in, wrong role" is `/` for now. If we want
  a dedicated `/forbidden` page later it's a one-line change.
- Scope expanded slightly to address the enum-extraction and 401
  feedback in the same pass — happy to split into a follow-up if it
  feels off-topic.
- Backend role middleware is unchanged; this PR only affects the
  client-side guard.

jasminerezai (changes requested):
Good job! Agree with Gunnar on both points. Fix the ActivitiesPage import in App.jsx and remove ROLES.LEADER from MANAGER_ROLES since the backend rejects it with a 403 anyway.

larrymijo:
> Good job! Agree with Gunnar on both points. Fix the ActivitiesPage import in App.jsx and remove ROLES.LEADER from MANAGER_ROLES since the backend rejects it with a 403 anyway.

- `ActivitiesPage.jsx` restored from main, so the default export is the
  list page again and the import in App.jsx now resolves correctly.
- `MANAGER_ROLES = [BOARD_MEMBER, ADMIN]` — LEADER moved to the new
  `EDITOR_ROLES = [LEADER, BOARD_MEMBER, ADMIN]`, which is what
  `/activities/:id/edit` uses. LEADER hitting `/activities/new` now
  redirects to `/` instead of getting a 403 on submit.

jasminerezai (approved):
Good Job Lawrence, looks good

gunnarwrld (approved):
Amazing job!

PR #79, opened 2026-05-20
MEMBER profile page with favorites and activity participation

### LINKED ISSUE #29: Profile page
Create a protected route at /profile that displays the logged-in user's information.

- Show user's name, email and role
- List their upcoming registered activities
- Show their saved favorite sports/activities with option to remove them

Note: The heart/favorite button on the Activities page (issue #21) 
feeds into this page. This page only handles the display and management 
of already saved favorites.

isaqelle:
## Summary

Implements the protected `/profile` page UI for authenticated users.

### Features

* Displays logged-in user's:

  * profile name
  * email
  * role
* Displays upcoming registered activities
* Displays saved favorite activities
* Allows users to remove favorites directly from the profile page
* Responsive card/grid layout for activities
* Added `ProfileService.js` for profile-related API fetching

### Integration

Uses the new:

```text id="zjlwm3"
/api/users/me
```

endpoint to fetch:

* profile data
* favorites
* participations

### Notes

* Upcoming activities are rendered from `participations`
* Favorites are rendered from `favorites`
* Uses existing `removeFavorite()` functionality from `FavoritesService`

### UI Improvements

* Redesigned profile page layout
* Added hero/profile section
* Converted activities into responsive card grids
* Added role badge styling

Closes #29

### DIFF
--- a/client/src/pages/ProfilePage.jsx
+++ b/client/src/pages/ProfilePage.jsx
@@ -16,7 +16,7 @@ export default function ProfilePage() {
 
   // ── State ───────────────────────────────────────────────
   // Stores general profile info:
-  // { name, email, role }
+  // { profileName, email, role }
   const [profile, setProfile] = useState({})
 
   // Favorite activities from backend
@@ -38,11 +38,11 @@ export default function ProfilePage() {
   //
   // The backend returns:
   // {
-  //   name,
+  //   profileName,
   //   email,
   //   role,
-  //   favorites: [],
-  //   participations: []
+  //   favorites: ActivityDto[],
+  //   participations: ScheduleDto[]
   // }
 
   useEffect(() => {
@@ -54,22 +54,20 @@ export default function ProfilePage() {
         const { data } =
           await fetchProfile(token)
 
-        // Favorites section
-        setFavoriteActivities(
-          data.favorites || []
-        )
 
-        // Upcoming activities section
+        const {
+          favorites,
+          participations,
+          ...profileData
+        } = data
+
+        setFavoriteActivities(favorites || [])
+
         setUpcomingActivities(
-          data.participations || []
+          participations || []
         )
 
-        // Remove nested arrays so profile only
-        // contains the actual user information.
-        delete data.favorites
-        delete data.participations
-
-        setProfile(data)
+        setProfile(profileData)
 
       } catch (error) {
 
@@ -157,7 +155,7 @@ export default function ProfilePage() {
                 marginBottom: '8px',
               }}
             >
-              {profile.name || 'Profile'}
+              {profile.profileName || 'Profile'}
             </h1>
 
             <p
@@ -220,10 +218,10 @@ export default function ProfilePage() {
             }}
           >
 
-            {upcomingActivities.map(activity => (
+            {upcomingActivities.map(schedule => (
 
               <Card
-                key={activity.id}
+                key={schedule.id}
                 padding="md"
                 shadow="sm"
               >
@@ -233,7 +231,7 @@ export default function ProfilePage() {
                     marginBottom: '8px',
                   }}
                 >
-                  {activity.name}
+                  {schedule.activity.name}
                 </h3>
 
                 <p
@@ -241,7 +239,7 @@ export default function ProfilePage() {
                     color: 'var(--color-text-muted)',
                   }}
                 >
-                  {activity.date}
+                  {new Date(schedule.startAt).toLocaleDateString()}
                 </p>
 
               </Card>
--- a/client/src/pages/ProfilePage.jsx
+++ b/client/src/pages/ProfilePage.jsx
@@ -8,16 +8,16 @@ import {
 } from '../services/ProfileService.js'
 import Card from '../components/ui/Card.jsx'
 import Button from '../components/ui/Button.jsx'
+import Badge, {
+  ROLE_VARIANT,
+} from '../components/ui/Badge.jsx'
 
 export default function ProfilePage() {
 
   // ── Auth ────────────────────────────────────────────────
-  const { token } = useAuth()
+  const { user, token } = useAuth()
 
   // ── State ───────────────────────────────────────────────
-  // Stores general profile info:
-  // { profileName, email, role }
-  const [profile, setProfile] = useState({})
 
   // Favorite activities from backend
   const [favoriteActivities, setFavoriteActivities]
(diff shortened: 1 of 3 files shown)


client/src/services/ProfileService.js
```diff
@@ -0,0 +1,19 @@
+import { API_BASE_URL } from './apiConfig.js'
+
+export async function fetchProfile(token) {
+
+  const response = await fetch(
+    `${API_BASE_URL}/api/users/me`,
```
larrymijo, line 6: heads up — i there is no /api/users/me , we might want to wait for it, or piggyback on the existing /api/auth/me instead of adding a second /me endpoint

client/src/pages/ProfilePage.jsx
```diff
@@ -0,0 +1,341 @@
...
+import {
+  fetchProfile,
+} from '../services/ProfileService.js'
+import Card from '../components/ui/Card.jsx'
+import Button from '../components/ui/Button.jsx'
+
+export default function ProfilePage() {
+
+  // ── Auth ────────────────────────────────────────────────
+  const { token } = useAuth()
```
larrymijo: we're already getting user = { id, email, name, role } from useAuth() — could we use that to fill in the hero section instantly instead of waiting on the fetch? then the API call is really just for favorites + participations. would also avoid the flash of "Loading profile..." every time

client/src/pages/ProfilePage.jsx
```diff
@@ -0,0 +1,341 @@
...
+  // The backend returns:
+  // {
+  //   profileName,
+  //   email,
+  //   role,
+  //   favorites: ActivityDto[],
+  //   participations: ScheduleDto[]
+  // }
+
+  useEffect(() => {
```
larrymijo, line 48: minor — should we guard against token being null here? on first mount during rehydration token can be null for a tick and we'd fire a request with Bearer null. something like if (!token) return at the top of loadProfile. ActivitiesPage does the same thing on line 50

JuriRappold (commented):
Looks good to me!  Just have to wait for #24 to merge (PR: #68 )

client/src/pages/ProfilePage.jsx
```diff
@@ -0,0 +1,341 @@
...
+      // Update UI immediately after removing
+      setFavoriteActivities(
+        favoriteActivities.filter(
+          activity => activity.id !== activityId
+        )
+      )
+
+    } catch (error) {
+
+      console.error(
```
larrymijo, line 104: right now if the remove fails the user just sees the card disappear and come back with no feedback. ActivitiesPage does an optimistic-with-rollback pattern on lines 91/120 — could we do the same here? just save the previous list and restore it on error

client/src/pages/ProfilePage.jsx
```diff
@@ -0,0 +1,341 @@
...
+          {/* User info */}
+          <div>
+
+            <h1
+              style={{
+                fontSize: '2.5rem',
+                marginBottom: '8px',
+              }}
+            >
+              {profile.profileName || 'Profile'}
```
larrymijo: we use name everywhere else (see AuthContext line 153 — email, name, role). is profileName intentional? if the backend hasn't been built yet maybe we align on name so it matches the existing user shape and the /auth/me response

client/src/pages/ProfilePage.jsx
```diff
@@ -0,0 +1,341 @@
...
+            <p
+              style={{
+                color: 'var(--color-text-muted)',
+                marginBottom: '12px',
+              }}
+            >
+              {profile.email}
+            </p>
+
+            {/* Role badge */}
```
larrymijo, line 201: we already have a <Badge> component for this check components/ui/Badge.jsx — it even has a ROLE_VARIANT map specifically for this and would also give us the right colors per role (MEMBER vs LEADER vs ADMIN) for free

client/src/pages/ProfilePage.jsx
```diff
@@ -0,0 +1,341 @@
...
+      console.error(
+        'Failed to remove favorite:',
+        error
+      )
+    }
+  }
+
+  // ── Loading State ───────────────────────────────────────
+  if (loading) {
+    return <p>Loading profile...</p>
```
larrymijo: small polish — could we wrap this in the same container/Card style as the rest of the page so it doesn't feel like a different page during the load? also no error state right now if the fetch fails — ActivitiesPage line 78 shows a simple error message we could mirror

client/src/pages/ProfilePage.jsx
```diff
@@ -0,0 +1,341 @@
...
+                    marginBottom: '16px',
+                    fontSize: '0.9rem',
+                  }}
+                >
+                  <strong>Location:</strong>{' '}
+                  {activity.location}
+                </p>
+
+                <Button
+                  variant="secondary"
```
larrymijo: Button doesn't have a secondary variant — it falls back to primary (line 143 in Button.jsx), so this is rendering as solid green. since it's a destructive action it'd make more sense as variant="danger", or variant="ghost" if we want something more subtle

larrymijo (changes requested):
Nice work on the layout 🙌 the card grid looks clean and the optimistic remove flow is good. Two things before we merge though: the /api/users/me endpoint doesn't exist on the backend yet so the page will 404 in its current state — we should sync with backend on that. And there's already a <Badge> component with a ROLE_VARIANT map made for exactly this use case, plus Button doesn't have a secondary variant so the Remove button is silently falling back to primary green. Left a few inline comments — most are small, but the badge + button variant ones are worth fixing before merge. Once those are in I think we're good 👌

jasminerezai (changes requested):
Good job! the /api/users/me dependency on Juri's unmerged PR #68 is the blocker. Everything else is polish that should also be fixed before merge.

client/src/pages/ProfilePage.jsx
```diff
@@ -0,0 +1,358 @@
...
+        setLoading(false)
+      }
+    }
+
+    loadProfile()
+
+  }, [token])
+
+  // ── Remove Favorite ─────────────────────────────────────
+  async function handleRemoveFavorite(activityId) {
```
jasminerezai, line 87: The remove currently waits for the API call before updating the UI, and if it fails there's no rollback or feedback. Use the optimistic pattern instead:

async function handleRemoveFavorite(activityId) {
  const previousFavorites = favoriteActivities
  setFavoriteActivities(favoriteActivities.filter(a => a.id !== activityId))
  try {
    await removeFavorite(activityId, token)
  } catch (error) {
    console.error('Failed to remove favorite:', error)
    setFavoriteActivities(previousFavorites)
  }
}

jasminerezai (commented):
Everything else looks good

jasminerezai (approved):
Good job!

larrymijo (approved):
nice job! approved

PR #87, opened 2026-05-20
30 error handling and loading states

### LINKED ISSUE #30: Error handling & loading states
Add across the app:
- Skeleton loaders for activity lists and detail pages
- Toast notifications for API errors
- 404 page for unknown routes
- Redirect to /login when JWT expires mid-session

larrymijo:
closes #30

adds the four pieces from the ticket — skeleton loaders, toast
notifications, a real 404 page, and proper handling for when the
jwt expires while the user is mid-session.

## what's in here

**toast system** — new `ToastContext` + `useToast` hook + `Toaster`
component in the ui library. replaces inline error states across
the app. styled with the existing design tokens (`--color-danger`,
`--color-success`, etc) so it looks like the rest of the project,
no new dependencies pulled in.

**skeleton loaders** — generic `Skeleton` primitive plus three
composed skeletons for the activity list, activity detail, and the
full schedule page. swaps in for the old `<p>Loading...</p>` text.
the schedule one keeps the whole page chrome visible while data
loads so the page doesn't pop in chunks. `ActivityDetailSkeleton`
is included for the next person to wire up (see notes).

**jwt expiry** — new `useAuthExpiredHandler` hook that wraps logout
+ toast + redirect to /login (preserving the current url so the
user lands back where they were after re-login). promoted from the
inline `handleAuthExpired` we already had in `ActivityFormPage`.
services now throw a typed `AuthExpiredError` on 401 so pages can
route session-expiry into the hook instead of treating it like any
other failure.

**404 page** — new `NotFoundPage` replaces the silent
`Navigate to="/"` catch-all in App.jsx. shows quick links and adapts
based on whether the user is logged in. the old behavior hid broken
internal links because every typo just bounced you to home.

also added the missing `@keyframes spin` to index.css — the auth
rehydration spinner in App.jsx was referencing it but it was never
defined anywhere.

## what to test

main things to actually exercise:

- reload `/` and `/activities` on slow 3g (devtools network
  throttling) to see the skeletons. shouldn't see `Loading...`
  text anywhere
- break the token (devtools → application → local storage → edit
  `hkif_token` to `broken`) while logged in, then click a heart
  icon on `/activities`. should toast "session expired" and
  redirect to /login. after logging in again you should land back
  on /activities, not /
- normal page refresh while logged in should still be silent — no
  toast on rehydration. this is the regression to watch for
- hit any made-up url like `/foo/bar`. should show the 404 page,
  NOT redirect to home (the url stays as `/foo/bar`)
- the existing favorites error banner on schedule page should
  still work as before (only shows up when "favorites only" is on
  AND favorites failed to load)

## notes

- `Toaster` sits inside `ToastProvider` which sits inside
  `AuthProvider` in App.jsx — order matters because
  `useAuthExpiredHandler` reads from both contexts
- `Skeleton` is `aria-hidden`, and the parent containers carry
  `aria-busy={loading}` so screen readers get told the page is
  loading without being read every grey rectangle


isaqelle:
The skeletons and the 404 page look awesome, love it!!

I noticed though that I can still favorite an activity with a broken token without getting logged out, it only happens after refreshing the page.

Since the PR notes mention refresh/rehydration should stay silent, I think something might still be triggering the auth-expired flow during startup/rehydration instead of only on actual expired sessions?

jasminerezai (approved):
Code looks good for this PR, good job!  Before merging, check what I replied to Isabelle's comment.

jasminerezai:
Good catch @isaqelle ! Lawrence, can you verify this manually?  log in, then go to DevTools → Application → Local Storage and change hkif_token to any random string, then click a heart icon. It should toast "session expired" and redirect to /login. If it doesn't, the bug is real and we should open a separate ticket for it since it's not introduced in this PR.

larrymijo:
> ### This branch has conflicts that must be resolved
> Use the [web editor](https://github.com/jasminerezai/hkif-web/pull/87/conflicts) or the command line to resolve conflicts before continuing.
> 
> * client/src/pages/ActivitiesPage.jsx
> * client/src/pages/SchedulePage.jsx

heya, reproduced it  — confirmed it's not introduced by this pr,
main has the same behavior. quick trace:

editing hkif_token in localStorage doesn't actually invalidate
the session — the AuthContext stores the jwt in React state when
you log in (separate from localStorage), and that's the copy
useAuth() returns to pages. so clicking a heart sends the real
in-memory token, the backend returns 200, and nothing 401s.

then on refresh, rehydration reads "broken" from localStorage,
the /me call fails, localStorage gets cleared, user logs out
silently. that part is working as intended.

to actually test the auth-expired flow this pr adds, you'd need
a real server-side 401. easiest way: temporarily set
JWT_EXPIRY=10s on the backend, log in, wait 15 secs, click a
heart — should toast + redirect. tried it and it works ✓

i suggest open a follow-up ticket for the localStorage-edit case — we
could have getAuthHeader read from localStorage at call-time
instead of from React state. small change but with implications
worth a separate discussion.

jasminerezai (approved):
Great job Lawrence!

isaqelle (approved):

PR #88, opened 2026-05-21
Bug participant count

### LINKED ISSUE #80: bug: participant count does not decrease on unregister
## Bug
The "spots left" counter displayed on activity/schedule cards does not update when a user unregisters from a session. The backend returns the updated `participantCount` in the unregister response but the client does not reflect it.

## Expected Behavior
- Spots left decreases by 1 when a user registers
- Spots left increases by 1 when a user unregisters
- Register button is disabled when `participantCount >= maxCapacity`

## Scope
- `client/src/pages/SchedulePage.jsx`  (re-fetch or update local state after unregister)
- `client/src/pages/ActivityDetailPage.jsx`(same)
- Disable the register button when the schedule is full

larrymijo:
## What
Fixes the bug where the "spots left" counter on schedule cards didn't move when a user registered or unregistered from a session.

The backend was already returning the updated `participantCount` in both `/participate` responses — the client just wasn't using it. Now it does.

## Changes
- New `client/src/services/ParticipationService.js` with `registerParticipation` and `unregisterParticipation`, following the same pattern as `FavoritesService.js`
- `SchedulePage.jsx`:
  - Activity shape now carries `maxCapacity` + `participantCount` instead of the misleading `availableSpots: maxCapacity`
  - The Attend/Leave button now actually hits the backend (it was a local-only toggle before)
  - Optimistic update with rollback on failure, same pattern as `handleToggleFavorite` in `ActivitiesPage`
  - Server's `participantCount` from the response overwrites the optimistic value, so the counter reflects the real count
  - Register button is disabled and shows "Full" when `participantCount >= maxCapacity`
- `ActivityDetailPage.jsx`: added a TODO comment for future register UI work. No functional change — this page doesn't have a register button yet, so there's nothing to bug-fix on it. Filed as a note for whoever picks up the next ticket.

## What to look for when testing
- Log in as a MEMBER, find a schedule card, click Attend → spots-left should drop by 1
- Click Leave → spots-left should go back up
- Open DevTools → Network tab → the response from `/participate` should match what's shown on the card
- If you can seed a schedule at full capacity, the button should show "Full" and be disabled. A user already attending should still be able to Leave.
- Mock-data mode (when backend is offline / no real schedules) should still work as a local toggle — no console errors and the button still flips between Attend/Leave

## Known limitation (follow-up ticket)
The very first paint of the page shows `maxCapacity` as the spots-left number, because `/api/schedules/current` doesn't return `participantCount` per schedule yet. It corrects itself the moment any user interacts with a card. I'll open a small backend follow-up ticket so the initial render is also accurate.

## Scope
Closes the SchedulePage part of the bug ticket. ActivityDetailPage is tracked separately via the TODO comment since adding a register flow there is more of a feature than a bugfix.

### DIFF
--- a/client/src/services/ParticipationService.js
+++ b/client/src/services/ParticipationService.js
@@ -1,0 +1,82 @@
+// All API calls for registering / unregistering a user
+// from a specific schedule live here.
+//
+// SchedulePage imports these — nothing else should call them directly.
+// Pattern mirrors FavoritesService.js so the codebase stays consistent.
+
+import { API_BASE_URL } from './apiConfig.js'
+
+// The participate endpoints are scoped under
+// /api/activities/:activityId/schedules/:scheduleId/participate
+//
+// We build the path inside each function rather than a single BASE
+// constant because both IDs are dynamic per call.
+
+// ── registerParticipation ─────────────────────────────────────
+// POST /api/activities/:activityId/schedules/:scheduleId/participate
+//
+// Backend returns: { status: 'success', data: { participantCount: number } }
+// We return the full JSON body so the caller can read participantCount
+// and update its local "spots left" counter accordingly.
+//
+// Protected — must send Bearer token.
+export async function registerParticipation(activityId, scheduleId, token) {
+
+  const response = await fetch(
+    `${API_BASE_URL}/api/activities/${activityId}/schedules/${scheduleId}/participate`,
+    {
+      method: 'POST',
+      headers: {
+        Authorization: `Bearer ${token}`,
+      },
+    }
+  )
+
+  if (!response.ok) {
+    // Surface the backend message when possible so the UI can show
+    // something more useful than a generic error. The backend uses
+    // ApiError which serialises to { status: 'error', message: '...' }.
+    let message = 'Failed to register for this session'
+    try {
+      const body = await response.json()
+      if (body?.message) message = body.message
+    } catch {
+      // Body wasn't JSON — fall back to the default message.
+    }
+    throw new Error(message)
+  }
+
+  return response.json()
+}
+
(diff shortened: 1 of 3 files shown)


client/src/services/ParticipationService.js
```diff
@@ -0,0 +1,82 @@
...
+  )
+
+  if (!response.ok) {
+    // Surface the backend message when possible so the UI can show
+    // something more useful than a generic error. The backend uses
+    // ApiError which serialises to { status: 'error', message: '...' }.
+    let message = 'Failed to register for this session'
+    try {
+      const body = await response.json()
+      if (body?.message) message = body.message
```
jasminerezai, line 42: the backend error shape uses body.error not body.message. Change to if (body?.error) message = body.error

client/src/services/ParticipationService.js
```diff
@@ -0,0 +1,82 @@
...
+        Authorization: `Bearer ${token}`,
+      },
+    }
+  )
+
+  if (!response.ok) {
+    let message = 'Failed to unregister from this session'
+    try {
+      const body = await response.json()
+      if (body?.message) message = body.message
```
jasminerezai, line 74: Same here, change to if (body?.error) message = body.error

jasminerezai (changes requested):
Good job Lawrence!!

client/src/services/ParticipationService.js
larrymijo: good catch, my bad — i looked at the controllers but didn't check errorHandler.ts. fixed in the latest commit

jasminerezai (approved):
Good job

Issue #31, opened 2026-05-04
Backend: Explicit Dto typing & missing return types of some queries


jasminerezai:
Validate all incoming request bodies using Zod (or express-validator).
Return clear 400 errors with field-level messages.
Cover: auth, activity creation/edit, participation, favorites. 

(This has been moved to Sprint 2. If it’s not done yet, please continue.)

JuriRappold:
What kind of Dto do we want? The explicit typing of `ActivityDto` or the one using the prisma types, like in `ScheduleDto`?

### ActivityDto - Explicit Typing
```typescript
export type Activity = {
  name: string,
  location: string,
  leaders: string[],
  description?: string | null,
  notes?: string | null,
  maxCapacity?: number | null,
  defaultStatus?: ActivityStatus,
  timeSlots: TimeSlot[],
};

export interface ActivityDto extends Omit<Activity, 'timeSlots' | 'leaders'> {
  leaders: {
    profileId: string;
    activityId: string;
  }[];
  timeSlots: {
    id: string;
    activityId: string;
    weekday: Weekday;
    startTime: Date;
    endTime: Date;
  }[];
}
```

### ScheduleDto - Implicit Typing with Prisma Types
```typescript
export type ScheduleDto = Schedule &{
    activity: ActivityTemplate,
    leaders: Pick<Profile, 'id' | 'profileName'>[]
};
```

JuriRappold:
Should I explicitly type the returns of some queries? Some are missing it rn: 

- READ.activityById
- READ.anyDaySchedule
- READ.participantsOf (unused function)
- READ.profilesFavorited (unused function)
- CREATE.newActivity
- UPDATE.updateActivity
- UPDATE.addTimeSlots
- DELETE.deleteActivity (not strictly necessary)
- DELETE.deleteSchedule (not strictly necessary)
- DELETE.unregisterParticipation (not strictly necessary)

JuriRappold:
If we think we are just gonna leave it as is (at least for now), i would close this issue @jasminerezai @gunnarwrld

jasminerezai:
Hey @JuriRappold! Good questions. On the DTO typing - let's go with explicit typing like ActivityDto. The ScheduleDto approach already bit us once with the leaders shape inconsistency that needed a whole PR to fix (#83), so let's not repeat that. Explicit typing keeps the API contract clean and decoupled from Prisma, and is easier for the frontend team to read without knowing the DB schema.


[...]

PR #89, opened 2026-05-21
chore: normalize leaders shape in ScheduleDto and ActivityDto

### LINKED ISSUE #83: chore: normalize `leaders` shape across ActivityDto and ScheduleDto
## Context
Found during PR #68 review. The `leaders` field has an inconsistent shape across DTOs:

```ts
// ActivityDto — raw Prisma join table row
leaders: { profileId: string; activityId: string }[]

// ScheduleDto — projected profile data
leaders: Pick<Profile, 'id' | 'profileName'>[]
```

One exposes the join table, the other exposes displayable profile data. Any component consuming both DTOs has to handle `leaders` differently depending on context.

## Problem
- Frontend cannot render leader names from `ActivityDto` without an additional lookup
- Inconsistency will cause confusion and bugs in conditional rendering (PR #82)
- `{ profileId, activityId }` leaks the internal Prisma join table shape into the API contract

## Proposed fix
Normalize both DTOs to use the same shape:

```ts
leaders: Pick<Profile, 'id' | 'profileName'>[]
```

This requires:
- [x] Update `ActivityDto` in `server/src/types/activity.types.ts`
- [x] Update `READ.activityById` and any other query that populates `leaders` on `ActivityDto` to include the profile projection
- [x] Update `CREATE.newActivity` and `UPDATE.updateActivity` if they return a shaped DTO
- [ ] Verify frontend consumers of `ActivityDto.leaders` are updated accordingly

JuriRappold:
Closes #83
## What changed:
- Changed `ActivityDto` to use the same shape for leaders as `ScheduleDto`
- adjusted the queries to return the correct shape to the controllers --> changing the query object and then formatting the new query object to the new Dto
    - in that scope i added a `formatActivity` function in `db/utils.ts`

## What didn't change :)
- no change required on the frontend, only the schedule page queried the leader name and that property didn't change


server/src/db/readQueries.ts
```diff
@@ -113,7 +113,9 @@ export class READ {
                 }
             }
         })
-        return favorites.map((a: { activity: ActivityDto }) => a.activity);
+        // return favorites.map((a: { activity: ActivityDto }) => a.activity);
```
jasminerezai: commented-out old line. Delete this line

server/src/db/readQueries.ts
```diff
@@ -142,23 +144,33 @@ export class READ {
     // all activities updated after a given timestamp
     // static async activitiesUpdatedAfter(lastRequest: Date){}
 
-    static async activityById(activityId: string): Promise<ActivityTemplate | null> {
+    static async activityById(activityId: string): Promise<ActivityTemplate | null> {//
```
jasminerezai: Can you fill in that // with a TODO? e.g. // TODO: normalize return type to ActivityDto. Just so we don't forget.

jasminerezai (approved):
Great job Juri!!

server/src/db/utils.ts
```diff
@@ -25,3 +25,20 @@ export function formatSchedule(schedule: any): ScheduleDto {
         leaders
     } satisfies ScheduleDto;
 }
+
+export function formatActivity(activity: any): ActivityDto {
+    let leaders: any = activity.leaders ?? [];
+    leaders = Array.isArray(leaders) ? leaders.map((el: { profile: { id: string; profileName: string | null } }) => el.profile) : [];
+    (activity as any).leaders = undefined;
```
gunnarwrld: This mutates the raw Prisma object directly, same issue that was flagged in PR #66 for the `formatSchedule` helper. The mutation is invisible to callers and bypasses TypeScript typing with `as any`.

The best approach is to destructure leaders out and spread the rest:
```
  export function formatActivity(activity: any): ActivityDto {
    const { leaders: rawLeaders, ...rest } = activity;
    const leaders = Array.isArray(rawLeaders) ? rawLeaders.map((el: { profile: { id: string; profileName: string | null } }) => el.profile) : [];
    return {
      ...rest,
      leaders,
    } satisfies ActivityDto;
  }
```
This avoids the as any cast and the mutation.

server/src/db/readQueries.ts
gunnarwrld: The return type comment is cut off but the declared type is still `ActivityTemplate`, not `ActivityDto`. This PR adds leaders include with profile projection, but the return type is never updated to reflect the new shape.

The raw Prisma result now has leaders as `{ profile: { id, profileName } }[]` but the return type says `ActivityTemplate` which has leaders as `LeaderActivity[]`. That's a type lie, callers get `ActivityTemplate` but actually receive a different shape.

server/src/db/utils.ts
```diff
@@ -25,3 +25,20 @@ export function formatSchedule(schedule: any): ScheduleDto {
         leaders
     } satisfies ScheduleDto;
 }
+
+export function formatActivity(activity: any): ActivityDto {
```
gunnarwrld, line 29: Maybe adding some type for this and `formatSchedule` functions?! (sorry im crazy about types)

gunnarwrld (commented):
Hey Juri! Really clean work here, the `formatActivity` utility follows the existing `formatSchedule` pattern well and the type change in `ActivityDto` is exactly what was needed.

gunnarwrld:
But i think we do need a change in the frontend. Since `ActivityDto.leaders` is now `Pick<Profile, 'id' | 'profileName'>[]`, the check in `ActivityDetailPage.jsx`:
```
  activity.leaders?.some(l => l.profileId === user.id)
```
will silently always return false, the field is now `l.id` not `l.profileId` no?

This needs to be updated to:
  `activityData.leaders?.some(l => l.id === user.id)`

Correct me if im wrong tho, not sure

JuriRappold:
> But i think we do need a change in the frontend. Since `ActivityDto.leaders` is now `Pick<Profile, 'id' | 'profileName'>[]`, the check in `ActivityDetailPage.jsx`:
> 
> ```
>   activity.leaders?.some(l => l.profileId === user.id)
> ```
> 
> will silently always return false, the field is now `l.id` not `l.profileId` no?
> 
> This needs to be updated to: `activityData.leaders?.some(l => l.id === user.id)`
> 
> Correct me if im wrong tho, not sure

@gunnarwrld I can't find the line of code in ActivityDetailsPage.jsx. Went through all `.jsx` files and *Ctrl + F* looked for the `leaders`  and `leader` keyword. and couldn't find an instance of using the id of the leader.

server/src/db/readQueries.ts
JuriRappold: yeah, we said that to be a separate issue, see the discussion [here](https://github.com/jasminerezai/hkif-web/issues/31#issuecomment-4506870316)
 Since it was never updated to return the ActivityDto in the first place

server/src/db/utils.ts
JuriRappold: good catch!

gunnarwrld:
@JuriRappold not here sorry, actually it's in Jasmine's PR #70 `ActivityDetailPage.jsx` line 78 uses `activityData.leaders?.some(l => l.profileId === user.id)`. So if your PR merges first, that check will break since leaders will now have id instead of profileId. Either Jasmine needs to update it in her PR, or you guys need to coordinate the order of merging.

gunnarwrld:
But she has a comment there acknowledging it needs to change to `l.id` once the shape is normalized so it should be fine, just letting you know!

server/src/db/utils.ts
JuriRappold: could add that, maybe in #31 ?

JuriRappold:
> @JuriRappold not here sorry, actually it's in Jasmine's PR #70 `ActivityDetailPage.jsx` line 78 uses `activityData.leaders?.some(l => l.profileId === user.id)`. So if your PR merges first, that check will break since leaders will now have id instead of profileId. Either Jasmine needs to update it in her PR, or you guys need to coordinate the order of merging.

ahh ok, will co-ordinate with her

server/src/db/utils.ts
JuriRappold: gonna leave as is for now

server/src/db/utils.ts
```diff
@@ -25,3 +25,22 @@ export function formatSchedule(schedule: any): ScheduleDto {
...
+    let leadersRaw = activity.leaders ?? [];
+    const leaders: leaderDto[] = Array.isArray(leadersRaw)
+        ? leadersRaw.map((el: { profile: { profileId: string; profileName: string | null } }) => el.profile)
+        : [];
+    const restOfActivity = {...activity}
+    delete restOfActivity.leaders;
+    return {
+        leaders: leaders,
+        timeSlots: restOfActivity.timeSlot,
```
jasminerezai: Typo - restOfActivity.timeSlot should be restOfActivity.timeSlots (missing the s). This will always return undefined and break any frontend reading time slots.

server/src/types/profile.types.ts
```diff
@@ -6,3 +6,8 @@ export type ProfileDto = Pick<Profile, 'profileName' | 'email' | 'role'> & {
     favorites: ActivityDto[],
     participations: ScheduleDto[]
 }
+
+export type leaderDto = {
+    profileId: string,
```
jasminerezai: Change profileId to id - the Prisma Profile shape uses id, not profileId. Using profileId here means l.profileId will always be undefined on the frontend.

jasminerezai (changes requested):

server/src/types/profile.types.ts
JuriRappold: ahh ok, i thought it was the other way around

gunnarwrld (approved):
Good job!!

PR #91, opened 2026-05-21
feat(server): add admin statistics endpoint (closes #84)

### LINKED ISSUE #84: Admin statistics endpoint
Backend GET endpoint returning aggregate statistics for the admin dashboard.

Should include:
- Total participants per activity
- Most popular activities
- Cancellation rates

Admin-only endpoint, restricted to ADMIN role.

gunnarwrld:
## Summary

Implements a backend `GET /api/admin/statistics` endpoint that returns aggregate dashboard statistics for administrators. This endpoint is protected and only accessible to users with the `ADMIN` role.

Closes #84

---

## Testing
Manually verified locally against seeded database:

Admin user successfully retrieves statistics with correct payload structure
Leader user correctly receives 403 Insufficient permissions
TypeScript compilation passes with no errors (npm run build)


jasminerezai (approved):
Hey Gunnar, great work on this! Really clean structure and the role guard is correct. Just one small thing before merging:

In `server/src/db/readQueries.ts` - `adminStatistics()` fetches all activities + all their schedules + all participation counts in one unbounded query. Totally fine for now with our dataset, but can you add a TODO comment on it so we don't forget post-submission?

Something like:
```ts
// TODO: This fetches all activities + schedules unbounded.
// Should be paginated or aggregated at DB level post-launch.
```

Just add this comment before merging

JuriRappold (approved):
looks good to me! Sry only now saw the PR

PR #94, opened 2026-05-22
Responsive design for the schedule page

### LINKED ISSUE #86: Responsive design
Make all pages responsive for mobile and tablet viewports.

Pages to cover:
- SchedulePage
- ActivitiesPage
- ActivityDetailPage
- ProfilePage
- ActivityFormPage

isaqelle:
### Changes made

* Added responsive calendar behavior using CSS Grid auto-fit/minmax
* Weekly view now collapses naturally on smaller screen sizes
* Improved mobile and tablet responsiveness for the schedule layout
* Added responsive styling for Weekly/Monthly toggle buttons
* Buttons now stretch only on smaller screens for better usability
* Cleaned up schedule toggle styling and removed old/conflicting responsive logic

### UI Improvements

* Better readability for activity cards on smaller screens
* More consistent spacing and responsive behavior
* Improved overall schedule page usability across devices

### Technical Notes

* Added `.schedule-toggle` responsive styles in `index.css`
* Updated calendar grid to use:
  `repeat(auto-fit, minmax(180px, 1fr))`
* Kept JSX cleaner by moving responsive behavior into CSS

### More notes
It seems that we already have working responsiveness for the other pages! I think I checked every page and it looks good, but let me know if I missed something.


larrymijo (changes requested):
hey! left a few thoughts before approving

The toggle refactor looks good, moving the inline styles into a class is much cleaner. The mobile button stretch is also a nice touch. Couple of things i wanted to raise though:

- the new repeat(auto-fit, minmax(180px, 1fr)) on the calendar grid kind of breaks the monthly view? a month is supposed to read as rows of weeks (7 cols), but with auto-fit the number of columns changes based on screen width — so weeks won't line up anymore. same issue in weekly view at medium widths, where you can end up with 5 days on one row and 2 on the next which looks a bit off. could we keep 7 columns by default and only collapse on smaller screens with a media query?

-the new media query uses 640px but the rest of the project uses 768px as the breakpoint — might be worth aligning so we don't end up with multiple breakpoints in different places

- small thing: gap: 12px could be gap: var(--space-3) to stay consistent with the design tokens. also missing a newline at the end of index.css 

other than that i think it's solid

jasminerezai (changes requested):
Good job Isabelle! Agreeing with Lawrence's points here. Once that's done I'll approve!

isaqelle:
good points! fixing it rn

larrymijo (approved):

jasminerezai (approved):

PR #95, opened 2026-05-22
feat(server): automated weekly schedule generation with cron

### LINKED ISSUE #56: feat(server): automated weekly schedule generation with cron
## What
Automatically generate schedule entries for the coming week every Monday morning using cron.

## Why
Currently schedules are seeded manually. This automates the process so the app stays up to date without any manual intervention.

## How
- Every Monday at 00:01, check if schedules already exist for the coming week
- If not, query all active ActivityTemplates with their timeSlots
- Generate Schedule rows for the week ahead (+ 1 week buffer)
- Reuse the `nextDateForWeekday()` logic already in the seed

## Dependencies
- Blocked by #55 (schedules endpoint type fixes should be merged first)

## Notes
- If the server is down on Monday, consider a startup check as a fallback
- Automation for past-week cleanup can be a follow-up issue

jasminerezai:
## What 
Automatically generates `Schedule` rows for the coming week every Monday at 00:01 using `node-cron`. Closes #56.

## The reason for implementing CRON:
Schedules were previously seeded manually. This automates the process so the app stays up to date without any manual intervention.

## Changes
- Installed `node-cron` + `@types/node-cron`
- `server/src/cron/scheduleGenerator.ts` - core generation logic:
  - Fetches all `ACTIVE` ActivityTemplates with their timeSlots
  - Computes the correct date for each slot in the coming week
  - Checks for existing schedules before creating (no duplicates possible)
  - Creates new `Schedule` rows with status `ACTIVE`
- `server/src/cron/index.ts` - cron job setup:
  - Registers the Monday 00:01 job
  - Runs a startup check on server boot as a fallback (in case the server was down on Monday)
- `server/src/server.ts` - wires `initCronJobs()` into the server startup

## What to test
- Start the server and check the console - you should see `[cron] Running startup schedule check...` followed by how many schedules were created or already existed
- Verify no duplicate schedules are created if you restart the server multiple times

image

## Notes
- Reuses `nextDateForWeekday()` logic from `seed.ts` and existing utils from `weekCalculator.ts`
- Past-week cleanup is out of scope and tracked as a follow-up

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -21,6 +21,7 @@
     "dotenv": "17.4.2",
     "express": "5.2.1",
     "jsonwebtoken": "9.0.3",
+    "node-cron": "^4.2.1",
     "pg": "8.20.0",
     "zod": "^4.4.3"
   },
@@ -30,6 +31,7 @@
     "@types/express": "5.0.6",
     "@types/jsonwebtoken": "9.0.10",
     "@types/node": "25.6.0",
+    "@types/node-cron": "^3.0.11",
     "@types/pg": "8.20.0",
     "nodemon": "3.1.14",
     "prisma": "7.8.0",
--- a/server/src/cron/index.ts
+++ b/server/src/cron/index.ts
@@ -1,6 +1,6 @@
 // server/src/cron/index.ts
 //
-// Sets up the cron job that generates schedules every Monday at 00:01.
+// Sets up the cron job that generates schedules every Monday at 00:05.
 // Also runs once on startup as a fallback in case the server was down
 // on Monday morning.
 
@@ -9,9 +9,9 @@ import { generateWeeklySchedules } from './scheduleGenerator.js';
 
 export function initCronJobs(): void {
     // Weekly schedule generation
-    // Runs every Monday at 00:01 server time.
-    // '1 0 * * 1' = minute 1, hour 0, any day of month, any month, Monday (1)
-    cron.schedule('1 0 * * 1', async () => {
+    // Runs every Monday at 00:05 server time.
+    // '5 0 * * 1' = minute 5, hour 0, any day of month, any month, Monday (1)
+    cron.schedule('5 0 * * 1', async () => {
         console.log('[cron] Running weekly schedule generation...');
         try {
             await generateWeeklySchedules();
@@ -29,5 +29,5 @@ export function initCronJobs(): void {
         console.error('[cron] Startup schedule check failed:', err);
     });
 
-    console.log('[cron] Weekly schedule generation cron job registered (Monday 00:01).');
+    console.log('[cron] Weekly schedule generation cron job registered (Monday 00:05).');
 }
--- a/server/src/cron/index.ts
+++ b/server/src/cron/index.ts
@@ -1,0 +1,33 @@
+// server/src/cron/index.ts
+//
+// Sets up the cron job that generates schedules every Monday at 00:01.
+// Also runs once on startup as a fallback in case the server was down
+// on Monday morning.
+
+import cron from 'node-cron';
+import { generateWeeklySchedules } from './scheduleGenerator.js';
+
+export function initCronJobs(): void {
+    // Weekly schedule generation
+    // Runs every Monday at 00:01 server time.
+    // '1 0 * * 1' = minute 1, hour 0, any day of month, any month, Monday (1)
+    cron.schedule('1 0 * * 1', async () => {
+        console.log('[cron] Running weekly schedule generation...');
+        try {
+            await generateWeeklySchedules();
+        } catch (err) {
+            console.error('[cron] Schedule generation failed:', err);
+        }
+    });
+
+    // Startup fallback
+    // Runs once when the server starts. Safe to call anytime —
+    // generateWeeklySchedules() checks for existing schedules
+    // before creating new ones, so no duplicates are possible.
+    console.log('[cron] Running startup schedule check...');
+    generateWeeklySchedules().catch(err => {
+        console.error('[cron] Startup schedule check failed:', err);
+    });
+
+    console.log('[cron] Weekly schedule generation cron job registered (Monday 00:01).');
+}
--- a/server/src/cron/scheduleGenerator.ts
+++ b/server/src/cron/scheduleGenerator.ts
@@ -83,7 +83,7 @@ export async function generateWeeklySchedules(): Promise<void> {
             const existing = await prisma.schedule.findFirst({
                 where: {
                     activityId: activity.id,
-                    startAt: { gte: startDay, lte: endDay },
+                    startAt,
                 },
             });
 
--- a/server/src/cron/scheduleGenerator.ts
+++ b/server/src/cron/scheduleGenerator.ts
@@ -1,0 +1,110 @@
+// server/src/cron/scheduleGenerator.ts
+//
+// Core logic for generating Schedule rows for the coming week.
+// Called by the cron job every Monday at 00:01, and once on
+// server startup as a fallback (in case the server was down on Monday).
+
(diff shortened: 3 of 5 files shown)


jasminerezai:
Noticed an edge case in the duplicate check after opening this - the original check looked for any schedule with the same `activityId` within the target week:

```ts
startAt: { gte: startDay, lte: endDay }
```

This would cause activities with multiple weekly time slots (e.g. Monday + Wednesday) to have their later slots skipped, since the first created schedule would trigger the exists check for all remaining slots that week.

Fixed in the latest commit - now checks against the exact `startAt` timestamp instead, so each slot is checked independently.

server/src/cron/index.ts
```diff
@@ -0,0 +1,33 @@
...
+// Also runs once on startup as a fallback in case the server was down
+// on Monday morning.
+
+import cron from 'node-cron';
+import { generateWeeklySchedules } from './scheduleGenerator.js';
+
+export function initCronJobs(): void {
+    // Weekly schedule generation
+    // Runs every Monday at 00:01 server time.
+    // '1 0 * * 1' = minute 1, hour 0, any day of month, any month, Monday (1)
```
JuriRappold: Is a minute after midnight enough of a difference to not interfere with the midnight bug?

jasminerezai: I could move it a few more minutes forward to be safer.  To 00:05 to give a bit more buffer around the midnight boundary? Or longer?

JuriRappold: yeah i think 5 min is safer

JuriRappold: but otherwise it looks good to me. btw is there a manual way to create a schedule?

jasminerezai: Okay, 'll move it to 00:05 for a bit more buffer. 
On the manual schedule creation - there's no dedicated endpoint for it right now, it runs via the CRON and seed. For testing you could temporarily change the cron expression to run every minute, or call generateWeeklySchedules() directly. What were you thinking - do you want a manual trigger endpoint?
@JuriRappold

server/src/cron/scheduleGenerator.ts
```diff
@@ -0,0 +1,110 @@
...
+    let skipped = 0;
+
+    for (const activity of activities) {
+        for (const slot of activity.timeSlots) {
+
+            // 2. Compute the exact date for this slot next week
+            const slotDate = nextDateForWeekday(startDay, slot.weekday);
+
+            const startAt = new Date(slotDate);
+            startAt.setUTCHours(
```
gunnarwrld, line 67: We should explicitly document that the DB storing times as UTC, just in case

server/src/cron/scheduleGenerator.ts
```diff
@@ -0,0 +1,110 @@
...
+            const endAt = new Date(slotDate);
+            endAt.setUTCHours(
+                slot.endTime.getUTCHours(),
+                slot.endTime.getUTCMinutes(),
+                0,
+                0,
+            );
+
+            // 3. Check if a schedule already exists for this slot
+            const existing = await prisma.schedule.findFirst({
```
gunnarwrld, line 89: If for some reason two time slots on the same activity share the same `startAt`
but have different `endAt` times (e.g. two back-to-back sessions starting at the
same hour), only the first will ever be created and the second will always be
skipped. Add `endAt` to the duplicate check:
```
  where: {
    activityId: activity.id,
    startAt,
    endAt,
  }
```

server/src/cron/index.ts
gunnarwrld: I agree with Juri here!

server/src/cron/scheduleGenerator.ts
```diff
@@ -0,0 +1,110 @@
...
+// Steps:
+//   1. Compute the date range for next week (Monday–Sunday)
+//   2. Fetch all ACTIVE ActivityTemplates with their timeSlots
+//   3. For each timeSlot, check if a schedule already exists in that range
+//   4. If not, create one with status ACTIVE
+//
+// Safe to call multiple times — the existence check prevents duplicates.
+export async function generateWeeklySchedules(): Promise<void> {
+    const now = new Date();
+    const nextMon = nextWeek(now, 1);              // base = one week ahead
```
gunnarwrld: I didn't quite understood the second argument '1', maybe add a brief comment explaining
the argument or rename it for clarity.

server/src/cron/scheduleGenerator.ts
```diff
@@ -0,0 +1,110 @@
...
+    const activities = await prisma.activityTemplate.findMany({
+        where: { defaultStatus: ActivityStatus.ACTIVE },
+        include: { timeSlots: true },
+    });
+
+    let created = 0;
+    let skipped = 0;
+
+    for (const activity of activities) {
+        for (const slot of activity.timeSlots) {
```
gunnarwrld, line 61: For each slot in the loop, a separate `prisma.schedule.findFirst` is
fired. If an activity has 5 time slots, that's 5 sequential round
trips per activity, and N activities × M slots = N*M queries total.

For a small club (like ours) this is fine. But if this ever runs over a large dataset,
maybe we could fetch all existing schedules for the week upfront:
```
  const existingSchedules = await prisma.schedule.findMany({
    where: {
      activityId: { in: activities.map(a => a.id) },
      startAt: { gte: startDay, lte: endDay },
    },
    select: { activityId: true, startAt: true, endAt: true }
  });
```
Then do an in-memory existence check inside the loop. Not a blocker for now
but worth noting as the activity count grows.

server/src/cron/index.ts
```diff
@@ -0,0 +1,33 @@
...
+    // Startup fallback
+    // Runs once when the server starts. Safe to call anytime —
+    // generateWeeklySchedules() checks for existing schedules
+    // before creating new ones, so no duplicates are possible.
+    console.log('[cron] Running startup schedule check...');
+    generateWeeklySchedules().catch(err => {
+        console.error('[cron] Startup schedule check failed:', err);
+    });
+
+    console.log('[cron] Weekly schedule generation cron job registered (Monday 00:01).');
```
gunnarwrld: This `console.log` still logs Monday 00:01 even though the timing is being updated to 00:05. Update the log message to match.

server/package.json
```diff
@@ -21,6 +21,7 @@
     "dotenv": "17.4.2",
     "express": "5.2.1",
     "jsonwebtoken": "9.0.3",
+    "node-cron": "^4.2.1",
```
gunnarwrld, line 24: pinning to a major version with ^ is fine, we just have to verify the Railway deployment picks up the right version on next deploy

gunnarwrld (approved):
Great job! Just the timing inconsistency (`startAt/endAt`) and Juri's suggestion and ready to go

server/src/cron/scheduleGenerator.ts
jasminerezai: 1 means "1 week ahead"

server/src/cron/scheduleGenerator.ts
jasminerezai: Good observation! Totally fine for our scale right now but I'll add a TODO comment so it doesn't get forgotten if the activity count ever grows.

server/package.json
jasminerezai: Noted. will keep an eye on it for the Railway deploy.

server/src/cron/scheduleGenerator.ts
jasminerezai: Great catch! Adding endAt to the check now.

server/src/cron/scheduleGenerator.ts
jasminerezai: Good point, adding a comment to make that explicit!

PR #96, opened 2026-05-22
implented bug fixes that should pass deployment (hopefully)

JuriRappold:
# Changes:
- changed `CREATE.newActivity` to properly return ActivityDto
- changed `UPDATE.updateActivity` to properly return ActivityDto
- added *id* property to ActivityDto
	- changed formatActivity accordingly

### DIFF
--- a/server/src/db/createQueries.ts
+++ b/server/src/db/createQueries.ts
@@ -43,7 +43,7 @@ export class CREATE {
         return formatActivity(activity);
     }
 
-    static async newActivity(newAct: Activity) {
+    static async newActivity(newAct: Activity): Promise<ActivityDto> {
         const activity = await prisma.activityTemplate.create({
             data: {
                 name: newAct.name,
@@ -70,11 +70,20 @@ export class CREATE {
                 },
             },
             include: {
-                leaders: true,
-                timeSlots: true
+                timeSlots: true,
+                leaders: {
+                    select: {
+                        profile: {
+                            select: {
+                                id: true,
+                                profileName: true
+                            }
+                        }
+                    }
+                }
             }
         });
-        return activity;
+        return formatActivity(activity);
     }
 
     static async newSchedule(data: {
--- a/server/src/db/readQueries.ts
+++ b/server/src/db/readQueries.ts
@@ -1,6 +1,6 @@
 import { prisma, ProfileRole } from "./prisma.js";
 import { startAndEndOfWeek } from "../utils/weekCalculator.js";
-import { ActivityTemplate, Profile } from "../generated/prisma/index.js";
+import { Profile } from "../generated/prisma/index.js";
 import { ApiError } from "../utils/ApiError.js";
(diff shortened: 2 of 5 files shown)


JuriRappold:
We can't see the `motivated-stillness/production` thingy in the PR, or?

JuriRappold:
forgot to update the `activities.controller.ts`

JuriRappold:
hmm only error its showing me there is this:
```typescript
export const newActivity = asyncHandler(
  async (req: Request<{}, {}, Activity>, res: Response<ApiResponse<ActivityDto>>) => {
    let newActivity: Activity;

    try {
      //here its complaining that id isn't part of the CreateActivitySchema
      newActivity = CreateActivitySchema.parse(req.body);
    } catch (error) {
      throw ApiError.badRequest(`Invalid request body: ${error}`);
    }
```

would change the declaration to: `let newActivity: Omit<Activity, 'id'>;` since a brand new Activity can't have an id yet

JuriRappold:
> hmm only error its showing me there is this:
> 
> ```ts
> export const newActivity = asyncHandler(
>   async (req: Request<{}, {}, Activity>, res: Response<ApiResponse<ActivityDto>>) => {
>     let newActivity: Activity;
> 
>     try {
>       //here its complaining that id isn't part of the CreateActivitySchema
>       newActivity = CreateActivitySchema.parse(req.body);
>     } catch (error) {
>       throw ApiError.badRequest(`Invalid request body: ${error}`);
>     }
> ```
> 
> would change the declaration to: `let newActivity: Omit<Activity, 'id'>;` since a brand new Activity can't have an id yet

same issue in the `Create.newActivity()` function

JuriRappold:
> > hmm only error its showing me there is this:
> > ```ts
> > export const newActivity = asyncHandler(
> >   async (req: Request<{}, {}, Activity>, res: Response<ApiResponse<ActivityDto>>) => {
> >     let newActivity: Activity;
> > 
> >     try {
> >       //here its complaining that id isn't part of the CreateActivitySchema
> >       newActivity = CreateActivitySchema.parse(req.body);
> >     } catch (error) {
> >       throw ApiError.badRequest(`Invalid request body: ${error}`);
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > would change the declaration to: `let newActivity: Omit<Activity, 'id'>;` since a brand new Activity can't have an id yet
> 
> same issue in the `Create.newActivity()` function

solved the issue in the latest commit

jasminerezai (approved):
Looks good, this should fix the deployment!

jasminerezai:
The timeSlots replace bug from #97 is in this file. 

 updateActivity calls deleteAllTimeSlots before addTimeSlots which removes existing slots. Fix is removing the deleteAllTimeSlots call so new slots are appended instead of replacing. One line change, worth adding here since you're already touching this file.

PR #98, opened 2026-05-22
Explicit Dto typing & Better query return values

### LINKED ISSUE #31: Backend: Explicit Dto typing & missing return types of some queries

JuriRappold:
Closes #31.

# Change in Dtos
- changed all Dtos to be of the explicit type (as described [here](https://github.com/jasminerezai/hkif-web/issues/31#issuecomment-4506505629)
    - note that I only did this with Dtos!
    - ProfileDto
    - ProfileSummaryDto
    - ScheduleDto
-  added `id` property for `ActivityDto (this is also done in #96 )

# Change in Queries
- changed READ.activityById, CREATE.newActivity, and UPDATE.updateActivity to return `ActivityDto`
    - queries themselves changed to get the required information 
    - formatting functions were used as well

# Change in `src/db/utils.ts`
- added the new `id` property
- moved leaders & timeSlots downwards towards the end of the object (aesthetic change)

### DIFF
--- a/client/src/pages/ActivityDetailPage.jsx
+++ b/client/src/pages/ActivityDetailPage.jsx
@@ -34,7 +34,6 @@ export default function ActivityDetailPage() {
 
         const result = await response.json()
         const activityData = result.data
-        console.error(activityData);
         setActivity(activityData)
 
         // Fetch participants in the same flow if the user is authorized (leader/admin)
--- a/client/src/pages/ActivityDetailPage.jsx
+++ b/client/src/pages/ActivityDetailPage.jsx
@@ -34,6 +34,7 @@ export default function ActivityDetailPage() {
 
         const result = await response.json()
         const activityData = result.data
+        console.error(activityData);
         setActivity(activityData)
 
         // Fetch participants in the same flow if the user is authorized (leader/admin)
@@ -41,7 +42,7 @@ export default function ActivityDetailPage() {
           const isUserLeader =
             user.role === 'ADMIN' ||
             user.role === 'BOARD_MEMBER' ||
-            (user.role === 'LEADER' && activityData.leaders?.some(l => l.profileId === user.id))
+            (user.role === 'LEADER' && activityData.leaders?.some(l => l.id === user.id))
 
           if (isUserLeader) {
             try {
@@ -75,9 +76,7 @@ export default function ActivityDetailPage() {
     return (
       user.role === 'ADMIN' ||
       user.role === 'BOARD_MEMBER' ||
-      // NOTE: ActivityDto.leaders uses { profileId, activityId } shape.
-      // If normalized to { id, profileName } (see issue #XX), change to l.id === user.id
-      (user.role === 'LEADER' && activity.leaders?.some(l => l.profileId === user.id))
+      (user.role === 'LEADER' && activity.leaders?.some(l => l.id === user.id))
     )
   }, [isAuthenticated, user, activity])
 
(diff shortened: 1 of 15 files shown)


jasminerezai (approved):
Clean work Juri! Explicit DTO typing is the right call and the l.id fix in ActivityDetailPage.jsx is exactly what was needed. One thing - make sure #96 merges first since both PRs touch the same files, otherwise you'll get conflicts here. Also noticed the activity: ActivityTemplate comment in ScheduleDto - good question, let's make that a post-submission tech debt ticket, too risky to change right before Monday.

Or, what do you guys think? @JuriRappold @gunnarwrld

JuriRappold:
Yeah will first merge the bug-fix branch #96. 
On the ActivityTemplate comment, i think it might too much. Becaues `ActivityDto` includes `timeSlots` and `leaders` again, that would mean we are sending that information twice in the same response.

jasminerezai:
Good point Juri, you're right!! changing it to ActivityDto would duplicate timeSlots and leaders in the response. Let's keep ActivityTemplate and just close that as a known design decision rather than tech debt.

client/src/pages/ActivityDetailPage.jsx
```diff
@@ -34,14 +34,15 @@ export default function ActivityDetailPage() {
 
         const result = await response.json()
         const activityData = result.data
+        console.error(activityData);
```
gunnarwrld: Remove this console error, it leaks the full activity payload

server/src/db/updateQueries.ts
```diff
@@ -34,16 +39,31 @@ export class UPDATE {
...
+                            select: {
+                                id: true,
+                                profileName: true
+                            }
+                        }
+                    }
+                }
             }
         });
+        return formatActivity(newAct);
```
gunnarwrld: This can return null, If the record was deleted between the update and the read (or if `activityId` is wrong), `newAct` is null and `formatActivity`(null) will throw an untyped runtime error. The return type `Promise<ActivityDto>` promises a non-null value but the code can violate that. Either throw `ApiError.notFound(...)` when `newAct` is null, or consolidate into the existing `prisma.activityTemplate.update(...)` call using `include` so you get the result in one query and avoid the race entirely.

server/src/types/schedule.types.ts
```diff
@@ -1,7 +1,16 @@
...
+
+export type ScheduleDto = {
+    id: string,
+    activityId: string,
+    startAt: Date,
+    endAt: Date | null,
+    status: ActivityStatus,
+    createdAt: Date,
+    updatedAt: Date,
+    activity: ActivityTemplate, // change to ActivityDto??
```
gunnarwrld: Yes, replace it to ActivityDto

gunnarwrld (changes requested):
Great work overall, just some small changes and ready to go!

client/src/pages/ActivityDetailPage.jsx
JuriRappold:
```suggestion
```

server/src/db/updateQueries.ts
JuriRappold:
```suggestion
if(!newAct) throw ApiError.notFound(`Activity to update couldn't be found`);
else return formatActivity(newAct);
```

server/src/types/schedule.types.ts
JuriRappold: But wouldn'tt it then carry duplicate data: leaders and (not strictly speaking) timeSlots?

PR #99, opened 2026-05-22
77 adminboard member activity management on profile page

### LINKED ISSUE #77: Admin/Board Member Activity Management on Profile Page
**Description**

Extend the `/profile` page for `ADMIN` and `BOARD_MEMBER` roles with additional activity management capabilities.

**Features**

* Display manageable activities
* Add quick-edit controls for activities
* Add edit icon/button on activity cards
* Navigate directly to activity edit form

**UI Ideas**

* Management dashboard layout
* Edit pen icon (✎)
* Admin action buttons on activity cards

**Notes**

This builds on the existing profile page and role-based UI structure.

Backend support for activity editing and role permissions already exists.

larrymijo:
## what this pr includes
- new "manage activities" section on `/profile`, only visible to ADMIN and BOARD_MEMBER
- each card has a ✎ pen in the top-right + an "Edit Activity" button, both jump straight to `/activities/:id/edit`
- swapped the old "Loading profile..." card for a proper page skeleton that mirrors the real layout (matches what we already do on ActivitiesPage with ActivityListSkeleton)

## files touched
- `client/src/services/ManageActivitiesService.js` — new, wraps `GET /api/activities`. kept in its own file so we can later swap to a scoped endpoint without touching the page
- `client/src/components/skeletons/ProfileSkeleton.jsx` — new, full-page skeleton + an exported `ManageActivitiesGridSkeleton` we reuse for the inner manage-loading state
- `client/src/pages/ProfilePage.jsx` — wires it all together: role gate, fetch effect, new section, skeleton

## how to test
- log in as ADMIN or BOARD_MEMBER → manage section should appear above upcoming / favorites, with the pen icon and edit button on every card
- log in as MEMBER → manage section should not appear at all (and no `/api/activities` call in the network tab)
- log in as LEADER → also no manage section (leader's own dashboard is a separate ticket #YY, on purpose here)
- click either the ✎ or the "Edit Activity" button → should land on `/activities/:id/edit` with the form prefilled
- throttle network to slow 3G in devtools → page skeleton should match the real layout, no shift when data arrives

## notes for review
- `GET /api/activities` is public + unscoped right now, so the page just pulls everything. for ADMIN/BOARD_MEMBER "manageable" = everything which is fine. if we later add per-role filtering, only the service file needs to change
- backend role checks on `PUT /api/activities/:id` are already in place, so even if a MEMBER somehow opened the edit URL by hand the server would block the save

### DIFF
--- a/client/src/components/skeletons/ProfileSkeleton.jsx
+++ b/client/src/components/skeletons/ProfileSkeleton.jsx
@@ -1,0 +1,242 @@
+// Skeleton placeholder for /profile.
+//
+// Mirrors the real ProfilePage layout so the page doesn't visually
+// jump when /api/users/me resolves. Sections rendered:
+//   1. Hero / user card     — always
+//   2. Manage Activities    — only if canManage (ADMIN / BOARD_MEMBER)
+//   3. Upcoming Activities  — always
+//   4. Favorite Activities  — always
+//
+// Why a single page-level skeleton instead of one-per-section:
+//   The hero + upcoming + favorites all land together when
+//   /api/users/me resolves, so splitting the file would just add
+//   import boilerplate without a real win.
+//
+// We also export ManageActivitiesGridSkeleton because the manage
+// list is fetched independently from the main profile call (see
+// ProfilePage's second useEffect), so the rest of the page can
+// already be visible while the management grid is still in flight.
+
+import React from 'react'
+import { Card, Skeleton } from '../ui'
+
+// ── Section heading placeholder ────────────────────────────────
+// Small helper so all three section titles look identical and we
+// only have to tweak one place if the typography changes.
+function SectionHeadingSkeleton() {
+  return (
+    <Skeleton
+      height="1.5rem"
+      width="35%"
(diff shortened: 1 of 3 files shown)


isaqelle (changes requested):
Looks very good! I like the little pen at the top corners, it's a nice touch. Also the skeleton looking fire as usual.

One thing we could include though is a “Create New Activity” button here that redirects to
`/activities/new`, since we don’t really have that flow exposed anywhere yet. Right now admins/board members can only edit existing activities.

jasminerezai (approved):
Hey Lawrence! Really clean PR. I like what Isabelle said

larrymijo:
> Looks very good! I like the little pen at the top corners, it's a nice touch. Also the skeleton looking fire as usual.
> 
> One thing we could include though is a “Create New Activity” button here that redirects to `/activities/new`, since we don’t really have that flow exposed anywhere yet. Right now admins/board members can only edit existing activities.

 — added a "create new activity"

isaqelle (approved):

PR #101, opened 2026-05-23
fix(server): resolve schedule query participation counts and date validation issues

gunnarwrld:
## What this does
Fixes Date Validation Crashes: Updates the `ScheduleDateSchema` validator to support full ISO datetime strings as well as short date strings (YYYY-MM-DD).

Corrects Participant Spot Counting: Adds database query inclusions and mapping logic to ensure that a schedule's participantCount accurately reflects the number of registered attendees in the database.

## Why this is necessary
Validation: The Zod validator previously used `z.iso.date()`, which only accepts short dates. If a client queried with a full ISO datetime string (which was also generated as the validator's own default value), the endpoint would crash with a validation error.

Incorrect Capacities: The client-side interface relies on `participantCount` to calculate available spots. Previously, the backend queries for creating, updating, and fetching a user's joined/profile schedules either omitted this count entirely or hardcoded it to 0. This resulted in the client showing incorrect spots remaining.

### DIFF
--- a/server/src/db/createQueries.ts
+++ b/server/src/db/createQueries.ts
@@ -113,6 +113,9 @@ export class CREATE {
                 status: finalStatus
             },
             include: {
+                _count: {
+                    select: { participations: true }
+                },
                 activity: {
                     include: {
                         leaders: {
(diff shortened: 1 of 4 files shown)


JuriRappold (approved):
Looks good to me! I think a few queries already have the `_count` property

jasminerezai (approved):
Looks good!

PR #104, opened 2026-05-23
feat: stats dashboard fully implemented.

### LINKED ISSUE #85: Admin statistics dashboard
Frontend admin-only page combining activity stats and management.

Should include:
- Charts using recharts showing participant counts and popular activities
- Uses data from the admin stats endpoint
- Gated to ADMIN role only

larrymijo:
## what was included in this pr
- new `/admin/statistics` page — admin-only dashboard with KPIs and recharts charts
- pulls aggregate data from `GET /api/admin/statistics` (already existed on the backend, was unexposed)
- "📊 Open Statistics" button added to the hero card on `/profile`, only visible to ADMIN
- full skeleton + error card loading states matching the pattern from `/profile`

- closes #85 —

## files touched
- `client/package.json` + `client/package-lock.json` — added `recharts` (sorry for the noisy lockfile diff 🙃)
- `client/src/services/AdminStatisticsService.js` — new, wraps `GET /api/admin/statistics` with token + `AuthExpiredError` handling (same pattern as `FavoritesService`)
- `client/src/components/skeletons/StatisticsSkeleton.jsx` — new, mirrors the real page layout so nothing jumps when data lands
- `client/src/pages/StatisticsPage.jsx` — new, the actual dashboard (KPIs + 3 charts + CSV export + refresh + last-updated stamp)
- `client/src/App.jsx` — new protected route `/admin/statistics` gated to `[ROLES.ADMIN]`, also pulls `ROLES` from `constants/roles.js`
- `client/src/pages/ProfilePage.jsx` — imports `ROLES`, adds `isAdmin` flag, renders the "Open Statistics" button inside the hero card

## what's on the page
- **KPI tiles** — Activities, Total Participants, Total Favorites, Cancellation Rate (with "X of Y schedules" hint)
- **Most Popular Activities** — horizontal grouped bar chart, top 10, shows participants vs favorites side by side
- **Participants per Activity** — vertical bar chart sorted descending, includes the long tail
- **Cancellation Rate per Activity** — horizontal red bar chart, hides activities with no schedules, sorted worst → best
- **Refresh button** — manually re-fetches without unmounting the page (existing charts stay visible while refreshing)
- **Last updated X ago** — live label that ticks every 30s, plus exact timestamp for the precise folks
- **Export CSV** — single merged CSV joining all three tables by activityId. Filename includes today's date so multiple exports don't collide

## ADMIN-only and not MANAGER_ROLES
The backend endpoint is `restrictToMinRole(ADMIN)`. If we let BOARD_MEMBER through on the frontend they'd just get a 403 → bad UX. Easier to keep the frontend gate strict and matching the backend. If we ever want to widen access, we change both sides at once.

## how to test
1. log in as ADMIN (`user1@example.com` / `user1pass` from seed)
2. go to `/profile` → "📊 Open Statistics" should be on the hero card → click it
3. should land on `/admin/statistics`, see skeleton briefly, then KPIs + 3 charts
4. wait ~40s → "Last updated" label should tick from "just now" → "30s ago" → "1 min ago"
5. click Refresh → button shows "Refreshing…", timestamp resets to "just now", charts stay visible (no full skeleton flash)
6. click Export CSV → downloads `hkif-statistics-YYYY-MM-DD.csv` with 6 columns
7. log out, log in as LEADER (`user2@example.com` / `user2pass`)
   - no stats button on `/profile`
   - typing `/admin/statistics` in the URL bar → bounced to `/`
8. same checks as MEMBER (`user6@example.com` / `user6pass`)
9. stop the backend → click Refresh → error card with "Try Again" appears → restart backend → "Try Again" recovers cleanly
10. devtools → 375px width → KPI tiles stack, header buttons wrap, no horizontal scrolling

## notes for review
- `recharts` is the only new dependency — lockfile diff is just that one install
- one `// eslint-disable-next-line no-unused-vars` on the `nowTick` state — it's intentional, the state value isn't read but flipping it forces a re-render so the relative-time label stays fresh. happy to refactor to a different approach if anyone has a cleaner suggestion
- CSV export is one merged file (not three) — easier to use in spreadsheets, but let me know if you'd rather have three separate exports
- the 30s tick interval is cleared on unmount in the effect — checked, no leak
- backend role middleware is still the real enforcement layer; the frontend gates just prevent doomed fetches and dead-end "no permission" screens

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -10,7 +10,8 @@
       "dependencies": {
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
-        "react-router-dom": "^6.22.0"
+        "react-router-dom": "^6.22.0",
+        "recharts": "^3.8.1"
       },
       "devDependencies": {
         "@vitejs/plugin-react": "^4.4.0",
@@ -775,6 +776,42 @@
         "@jridgewell/sourcemap-codec": "^1.4.14"
       }
     },
+    "node_modules/@reduxjs/toolkit": {
+      "version": "2.12.0",
+      "resolved": "https://registry.npmjs.org/@reduxjs/toolkit/-/toolkit-2.12.0.tgz",
+      "integrity": "sha512-KiT+RzZbp6mQET+Mg+h2c97+9j1sNflUxQkIHI7Yuzf6Peu+OYpmkn6nbHWmLLWj+1ZODUJFwGZ7gx3L9R9EOw==",
+      "license": "MIT",
+      "dependencies": {
+        "@standard-schema/spec": "^1.0.0",
+        "@standard-schema/utils": "^0.3.0",
+        "immer": "^11.0.0",
+        "redux": "^5.0.1",
+        "redux-thunk": "^3.1.0",
+        "reselect": "^5.1.0"
+      },
+      "peerDependencies": {
+        "react": "^16.9.0 || ^17.0.0 || ^18 || ^19",
+        "react-redux": "^7.2.1 || ^8.1.3 || ^9.0.0"
+      },
+      "peerDependenciesMeta": {
+        "react": {
+          "optional": true
+        },
+        "react-redux": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/@reduxjs/toolkit/node_modules/immer": {
+      "version": "11.1.8",
+      "resolved": "https://registry.npmjs.org/immer/-/immer-11.1.8.tgz",
+      "integrity": "sha512-/tbkHMW7y10Lx6i1crLjD4/OhNkRG+Fo7byZHtah0547nIeXYcpIXaUh0IAQY6gO5459qpGGYapcEOHtFXkIuA==",
+      "license": "MIT",
+      "funding": {
+        "type": "opencollective",
+        "url": "https://opencollective.com/immer"
+      }
+    },
     "node_modules/@remix-run/router": {
       "version": "1.23.2",
       "resolved": "https://registry.npmjs.org/@remix-run/router/-/router-1.23.2.tgz",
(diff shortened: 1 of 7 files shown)


jasminerezai (approved):
Great work Lawrence! Really clean implementation. 

After submission, when you get the chance, StatisticsPage.jsx is 887 lines total which is quite large. I think its worth splitting into smaller chart subcomponents (one per chart). Not a blocker at all for now! just a note for later. Good job again

gunnarwrld (approved):
Overall this looks really good, the feature lines up nicely with #85 and the UX details (skeletons, error card, refresh, CSV export, admin-only gating) are super well thought-through. 🙌

A few comments, mostly around maintainability:

1. `StatisticsPage.jsx` is currently ~887 lines. As Jasmine suggested, I’d also strongly suggest a follow-up refactor that keeps this as the container (data fetch + layout) and extracts each chart into its own component (e.g. `StatisticsPopularActivitiesChart`, `StatisticsParticipantsPerActivityChart`, `StatisticsCancellationRateChart`). That will make future changes much easier.

2. For the `nowTick` state that only exists to force re-renders for the “Last updated X ago” label: instead of disabling `no-unused-vars`, I’d prefer either a small `useNowTick` hook or some explicit usage of the state value so the intent is clearer and we don’t need an eslint-disable.

3. Role gating looks correct (frontend matches `restrictToMinRole(ADMIN)`), but it might be worth centralising the `isAdmin` logic in a small helper so we don’t duplicate role checks across pages if we ever widen access.

4. CSV export: merging into a single 6-column file is a nice choice. Could be good to add a short comment/JSDoc on the CSV builder describing the columns so future changes don’t accidentally break the format.

None of these are blockers for me, happy for this to go in as-is and we can open a small follow-up for splitting `StatisticsPage.jsx` into subcomponents.

isaqelle (approved):
Agreed with what Jasmine and Gunnar said. Great job as always Lawrence!! Super nice tables

PR #109, opened 2026-05-24
refactor — statistics page 

### LINKED ISSUE #108: refractor of statistics page
follow-up ticket for splitting StatisticsPage.jsx into subcomponents.

- container (data fetch + layout) and extract each chart into its own component (e.g. StatisticsPopularActivitiesChart, StatisticsParticipantsPerActivityChart, StatisticsCancellationRateChart). That will make future changes much easier.

- also look into nowTick state that only exists to force re-renders for the “Last updated X ago” label: instead of disabling no-unused-vars, I’d prefer either a small useNowTick hook or some explicit usage of the state value so the intent is clearer and we don’t need an eslint-disable.
- centralising the isAdmin logic in a small helper so we don’t duplicate role checks across pages if we ever widen access.
-add a short comment/JSDoc on the CSV builder describing the columns so future changes don’t accidentally break the format.

larrymijo:
## Wts included in this pr
follow-up refactor on `StatisticsPage.jsx` — no user-facing changes, just splitting the page into smaller pieces and cleaning up two code suggestionsthat came up in PR-85 review.

- split the chart rendering into three separate components (`StatisticsPopularActivitiesChart`, `StatisticsParticipantsPerActivityChart`, `StatisticsCancellationRateChart`), each in its own file
- extracted the KPI tiles into `StatisticsKpiTiles` and the chart card wrapper into `ChartCard`
- moved the chart color tokens into a shared `chartTokens.js`
- replaced the `nowTick` useState + `eslint-disable no-unused-vars` with a new `useNowTick` hook
- centralised the `user?.role === ROLES.ADMIN` check into a `useIsAdmin` hook
- added proper JSDoc on `buildCsv` documenting the column contract

this PR is purely the refactor to address all of that. no behavior changes, no new features.

## files touched
new files:
- `client/src/hooks/useNowTick.js` — forces a re-render every N ms for relative-time labels. internal `useState` is `[, setTick]` so there's nothing to lint-disable
- `client/src/hooks/useIsAdmin.js` — single source of truth for "is current user ADMIN", returns boolean
- `client/src/components/statistics/chartTokens.js` — named color constants (all CSS vars from index.css, no new colors)
- `client/src/components/statistics/ChartCard.jsx` — shared title + description + card shell
- `client/src/components/statistics/StatisticsKpiTiles.jsx` — the four KPI tiles + their grid + the totals calculation
- `client/src/components/statistics/StatisticsPopularActivitiesChart.jsx` — top-10 horizontal grouped bar chart
- `client/src/components/statistics/StatisticsParticipantsPerActivityChart.jsx` — vertical bar chart, all activities
- `client/src/components/statistics/StatisticsCancellationRateChart.jsx` — red horizontal bar chart

modified:
- `client/src/pages/StatisticsPage.jsx` 
- `client/src/pages/ProfilePage.jsx`

## how to test
since this is a pure refactor, the goal is to confirm nothing broke. all the same checks from PR-85 should still pass:

1. log in as ADMIN (`user1@example.com` / `user1pass`)
2. `/profile` → "📊 Open Statistics" button still on the hero card → click it
3. `/admin/statistics` → skeleton briefly → same KPI tiles + same three charts as before (visually identical)
4. wait ~40s → "Last updated" label should tick ("just now" → "30s ago" → "1 min ago"). this is the useNowTick hook working
5. refresh button → still shows "Refreshing…", timestamp resets, charts stay visible
6. export CSV → downloads same 6-column file as before
7. as LEADER / MEMBER → no stats button on `/profile`, `/admin/statistics` URL bounces to `/`
8. devtools console → no warnings about unused vars

## notes for review
- the JSDoc on `buildCsv` is mostly a warning to future-us: the columns + their order are part of an unwritten contract with whoever's consuming the CSV. comment makes that explicit so it doesn't get silently broken in a future PR
- naming: i went with the explicit `StatisticsXyzChart` prefix the ticket suggested, even though they're in a `statistics/` folder. the folder + prefix combo makes them very findable via filename search, and the prefix is self-documenting if a component ever gets imported elsewhere
- ActivityDetailPage still has inline `user.role === 'ADMIN'` checks in a couple of places. didn't touch them in this PR since the ticket is specifically about cleanup around StatisticsPage, but they could become `useIsAdmin()` calls in a small follow-up if we want. let me know if you'd prefer i fold that in here
- `chartTokens.js` only has color constants for now. if other chart-specific shared values come up later (default heights, margin presets, etc.) they should land here too

### DIFF
--- a/client/src/components/statistics/ChartCard.jsx
+++ b/client/src/components/statistics/ChartCard.jsx
@@ -1,0 +1,56 @@
+// ChartCard — wraps a recharts chart in a consistent card shell
+// with a title and description.
+//
+// Why this lives in its own file:
+//   All three statistics charts use the exact same title +
+//   description + spacing treatment. Inlining the markup in each
+//   chart component would mean any visual tweak (e.g. tightening
+//   the description margin) needs to be made in three places.
+//
+// Layout assumptions:
+//   - children is the chart itself (a recharts <ResponsiveContainer>
+//     or an empty-state <p>)
+//   - card padding="lg" matches the existing /profile cards so the
+//     statistics page feels visually consistent with profile
+
+import React from 'react'
+import Card from '../ui/Card.jsx'
+import { CHART_COLOR_TEXT_MUTED } from './chartTokens.js'
+
+export default function ChartCard({
+  title,
+  description,
+  children,
+}) {
+  return (
+
+    <Card
+      padding="lg"
+      shadow="sm"
+      style={{ marginBottom: '24px' }}
+    >
+
+      <h2
+        style={{
+          fontSize:     '1.2rem',
+          marginBottom: '4px',
+        }}
+      >
+        {title}
+      </h2>
+
+      <p
+        style={{
+          fontSize:     '0.9rem',
+          color:        CHART_COLOR_TEXT_MUTED,
+          marginBottom: '20px',
+        }}
+      >
+        {description}
+      </p>
+
+      {children}
+
+    </Card>
+  )
+}
--- a/client/src/components/statistics/StatisticsCancellationRateChart.jsx
+++ b/client/src/components/statistics/StatisticsCancellationRateChart.jsx
@@ -1,0 +1,133 @@
+// StatisticsCancellationRateChart — horizontal red bar chart
+// showing cancellation rate (%) per activity, sorted worst-first.
+//
+// Red bars (danger token) signal "higher = bigger problem" so the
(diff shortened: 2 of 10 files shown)


isaqelle (approved):
Great job Lawrence! Looks way cleaner now. Tested the stats page and seems to be fully working.

jasminerezai (approved):
Good

PR #110, opened 2026-05-24
Leader profile dashboard

### LINKED ISSUE #76: Leader Profile Dashboard
**Description**

Extend the /profile page for users with the LEADER role by adding management-focused activity sections and controls.

**Features**

- Display activities the leader is responsible for

- Show current activity status (ACTIVE, CANCELLED)

- Show participant count

- Display upcoming session/date information

**UI Ideas**

- Dashboard-style activity cards

- Status badges

- “Change Status” button/action

- Notes

**Backend support for:**

leader role
leading activities
status updates

already exists.

This issue only covers frontend UI + integration.

isaqelle:
## Summary

Extends the leader experience on `/profile` and activity management flows.

### Added

* Leader dashboard section on `/profile`
* Display of activities assigned to the logged-in leader
* "Manage" button for assigned activities
* Assigned leaders can now access the existing activity edit page
* Leaders can now update:

  * location
  * description
  * notes
  * default activity status

### Updated

* Added leader visibility for the "Edit Activity" button on `ActivityDetailPage`
* Reused existing `ActivityFormPage` for leader management flow
* Added role-based field restrictions in `ActivityFormPage`

### Restrictions

Leaders cannot edit:

* activity name
* max capacity
* assigned leaders
* add/remove/edit time slots

### UI Improvements

* Added status badges to leader activity cards
* Improved dashboard consistency/styling
* Redirect leaders back to `/profile` after saving changes


larrymijo (changes requested):
## review notes

nice work overall, leader flow works and reusing `ActivityFormPage` was a smart call. just a few things before merging

**1. inline fetch in `ProfilePage.jsx` instead of a service**

we have `ProfileService`, `FavoritesService`, `ManageActivitiesService` already, the new leader fetch sits inline in the component (around lines 184-208). would be cleaner to put it behind a service like the rest, the comment on `ManageActivitiesService.js` actually anticipates this


**2. missing loading / empty / error states on the leader section**

the manage activities section above it has all three (skeleton, empty card, error log). the leader section just renders nothing while loading and nothing if the array is empty. a new leader with no assignments would just see an empty header. Could be added in a follow up ticket tho

**3. no 401 handling on the leader fetch**

rest of the app uses `useAuthExpiredHandler`. if the token expires mid session this just silently fails, would be good to match the pattern

**4. status badge always shows on leader cards**

on the manage cards above the badge only renders when status is `CANCELLED`, on the new leader cards every activity gets a green `ACTIVE` badge. small thing but worth matching

**small stuff**

- disabled fields in `ActivityFormPage` dont tell the leader why theyre disabled, a small helper text would help
- capitalization on the save button is inconsistent, "Save Changes" vs "Save changes"
- `isLeader` check is duplicated in `ActivityDetailPage` (useEffect line 42 and render line 75), could pull it into a `useMemo`

overall really solid pr, once the service refactor and the missing states are in i think this is good to merge

larrymijo (approved):
nice! review points adressed, approved!

just a note - loading state for the leader section still missing, has no loading flag at all so during the fetch it briefly shows the empty state "No assigned activities yet" before the cards pop in, which is a bit misleading. would be good to add a leaderLoading flag and a small skeleton or just hide the empty state until the fetch resolves. Nothing importat but worth mentioning for a future improvement.

jasminerezai (approved):

jasminerezai:
Really nice work on the leader flow and reusing ActivityFormPage was a smart call.

One thing to flag - in LeaderActivitiesService.js the fetch hits /api/activities and filters client-side. This works for now but the backend doesn't have a filtered endpoint yet, so I've opened a ticket for that. Once it's done the service just needs a small update to pass the leaderId as a query param instead.

PR #112, opened 2026-05-25
fix: only run startup schedule check on Mondays

jasminerezai:
fix - CRON startup duplicate guard

the startup fallback in initCronJobs() was running on every server 
restart regardless of day. on Mondays this caused the startup fallback 
and the 00:05 cron to both fire, inserting duplicate Schedule rows.

fix: wrapped the startup call in a getDay() === 1 check so it only 
runs if today is actually Monday.

note: duplicate rows from today still need to be manually cleaned 
from the db - this PR only prevents it happening again.

### DIFF
--- a/server/src/cron/index.ts
+++ b/server/src/cron/index.ts
@@ -1,8 +1,8 @@
 // server/src/cron/index.ts
 //
 // Sets up the cron job that generates schedules every Monday at 00:05.
-// Also runs once on startup as a fallback in case the server was down
-// on Monday morning.
+// Also runs once on startup IF today is Monday, as a fallback in case
+// the server was down at 00:05.
 
 import cron from 'node-cron';
 import { generateWeeklySchedules } from './scheduleGenerator.js';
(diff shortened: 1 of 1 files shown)


server/src/cron/index.ts
gunnarwrld, line 1: I was thinking: if the cron fires at 00:05 and the server restarts within that same Monday morning window, the guard still allows both to run concurrently. 

Maybe the real fix is ensuring  `generateWeeklySchedules()` is idempotent and transactionally safe, not just skipping it outside Mondays. 

The DB-level deduplication guard inside  `generateWeeklySchedules()` should be the primary defense; although this `getDay()`  check is a reasonable secondary guard.

With that said, this fix is safe to merge as a hotfix given the duplicate rows are actively occurring. The major idempotency issue should be tracked as a follow-up.

jasminerezai: Hey @gunnarwrld, good point! Just to clarify, the idempotency guard is already in place in generateWeeklySchedules(). Before creating any schedule it checks if one already exists for that (activityId, startAt, endAt) combination and skips it if so, so duplicate runs are safe. The getDay() check is just a secondary guard on top of that as you mentioned.

I'll open a ticket after our presentation to add a proper unique constraint at the DB level as the final fix. Does that work for you?

JuriRappold (approved):
Looks good to me!

PR #113, opened 2026-05-25
feat(activities): add leaderId filter and update leader activities client

### LINKED ISSUE #111: Add leaderId filter to GET /api/activities
Currently GET /api/activities returns all activities and the frontend filters client-side in LeaderActivitiesService.js. This means every leader downloads the full activity list just to show their assigned activities.
The backend should support a query parameter like ?leaderId=123 so only the relevant activities are returned. Once this is in, LeaderActivitiesService.js can be updated to pass the leaderId in the request instead of filtering client-side.
Acceptance criteria:

GET /api/activities?leaderId=:id returns only activities where the given profile is assigned as a leader
LeaderActivitiesService.js is updated to use the new query param

gunnarwrld:
## What it does

This PR adds support for filtering activities by `leaderId` in the backend and updates the leader activities client to use the new query parameter.

### Backend
- Extend `GET /api/activities` to accept an optional `leaderId` query param.
- Pass `leaderId` from the controller into `READ.allActivities(leaderId)`.
- Update `READ.allActivities` to optionally filter activities by leaders matching the given `leaderId`.

### Frontend
- Update `LeaderActivitiesService.js` to request `/api/activities?leaderId=${userId}` instead of the unfiltered endpoint.
- Remove client-side filtering of activities by leader and return the server-filtered list directly.
### Notes
- Keeps existing behavior when no `leaderId` is provided.
- Closes #111.

### DIFF
--- a/client/src/services/LeaderActivitiesService.js
+++ b/client/src/services/LeaderActivitiesService.js
@@ -6,7 +6,7 @@ export async function fetchLeaderActivities(
   handleAuthExpired
 ) {
   const res = await fetch(
-    `${API_BASE_URL}/api/activities`,
+    `${API_BASE_URL}/api/activities?leaderId=${userId}`,
     {
       headers: {
         ...getAuthHeader(),
@@ -27,9 +27,5 @@ export async function fetchLeaderActivities(
     )
   }
 
-  return json.data.filter(activity =>
-    activity.leaders?.some(
-      leader => leader.id === userId
-    )
-  )
+  return json.data
 }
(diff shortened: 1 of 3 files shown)


jasminerezai (approved):
Looks good! just add the leaderId input validation in the controller before merge. it's not a blocker. The worst case is an empty response, not a crash or data leak.

gunnarwrld:
> Looks good! just add the leaderId input validation in the controller before merge. it's not a blocker. The worst case is an empty response, not a crash or data leak.

Oh thanks for spotting that, Jasmine, I completely missed it. I’ll update it rn.

Issue #114, opened 2026-05-25
Add unique constraint on (activityId, startAt) to Schedule table


jasminerezai:
dd a DB-level guard to make duplicate schedule rows impossible, even in race conditions where the cron fires and the server restarts in the same window.

Tasks:
Add @@unique([activityId, startAt]) to the Schedule model in schema.prisma
Run npx prisma migrate dev --name add_unique_schedule_constraint
Handle the constraint violation error gracefully in generateWeeklySchedules() so the cron job doesn't crash

JuriRappold:
Would potentially also add a unique constraint to the activity name to prevent duplicate activityTemplates

JuriRappold:
It could also be easier to do this in the TimeSlot table: `@@unique([weekday, startTime])`

PR #117, opened 2026-05-25
Prisma Client Error Handling, Bug fixes and 409-response for POST newActivity

### LINKED ISSUE #107: bug(db): Prisma foreign key constraint errors return unhandled 500
## Description

Some database-level errors (e.g. Prisma foreign key constraint violations) are currently bubbling up as unhandled exceptions and appear as generic 500 responses.

We should map these to proper `ApiErrors` with meaningful status codes and messages, instead of leaking raw Prisma errors.

## Examples / scenarios

- Creating or updating a record with references to non-existent related entities (FK violation).
- Deleting records that are still referenced by other tables, where the relation is not set to cascade.

Currently behaviour:

- Response status: `500 Internal Server Error`
- Response body: generic 500
- Logs: raw Prisma error (e.g. `P2003` foreign key constraint failed).

## Expected behaviour

- These errors should be caught in the global error handler and translated into appropriate HTTP errors, for example:
  - `400 Bad Request` or `409 Conflict` for FK constraint violations.
- Response body should use the same structured error shape as `ApiError`, e.g.:

  ```json
  {
    "statusCode": 409,
    "error": "Foreign key constraint failed",
    "details": "Cannot delete activity that still has participants"
  }
  ```

## Proposed fix / pointers

- In the global error handler, inspect Prisma errors by:
  - `instanceof Prisma.PrismaClientKnownRequestError`, and
  - `error.code` (e.g. `P2003` for foreign key constraint failed).
- Map known Prisma error codes to appropriate `ApiError` instances.
- Ensure unknown Prisma errors still surface as 500, but without leaking internal details.

## Testing

Add Vitest + Supertest integration tests that:

- Trigger a FK violation (e.g. deleting or creating with invalid relation).
- Assert that:
  - The response status is `409` (or chosen status).
  - The response body contains a clear, user-friendly error.
  - No raw Prisma error details are exposed.

## Ownership

- Juri to implement Prisma error handling on Monday.
- Gunnar to add regression tests for these cases in the backend test suite.

## Links

- Parent issue: #32 (End-to-end testing & bug fixes)
- Related testing issue: #102 / #105

JuriRappold:
Closes #107. 
---
# Failed Prisma Queries
- The prisma errors just bubble up to the global error handler, where there is an if-statement checking if its a prisma client error, which pretty much means a bad input, therefore sending a 400-response. Any other prisma errors (like the connection to DB failing) are still unhandled.
```typescript
// in errorHandler.ts
if( err instanceof PrismaClientKnownRequestError ){
	const prismaErr = prismaErrorFormatter(err);
	res.status(400).json({
	  error: prismaErr,
	  statusCode: 400,
	})
}
```

*To format the prisma errors (to get the essential information), I used the following function:*
```typescript
export function prismaErrorFormatter(error: PrismaClientKnownRequestError) {
    const driverAdaptErr = error.meta?.driverAdapterError as any;
    const simpleErr =  {
        code: error.code,
        name: driverAdaptErr.cause.kind,
        fields: driverAdaptErr.cause.constraint.fields,
        table: error.meta?.modelName,
        message: driverAdaptErr.cause.originalMessage
    }
    return simpleErr;
}
```

---

# Bugs fixed

## Where `checkTimeFormat()` failed: (BE)

- checkTimeFormat() would fail, when the individual time slots (HH, mm, or ss) would pass the number check. This allowed times like these:
	- *18:6:00* --> as 6 is an integer and below 59
- solution: replaced with reggex check, that checks for both existing numbers

### old function
```typescript
const timeSplit = time.split(":");
    if (timeSplit.length !== 3) return false;
    const [hours, minutes, seconds] = timeSplit.map(Number);
    return (
        Number.isInteger(hours) && hours! >= 0 && hours! <= 23 &&
        Number.isInteger(minutes) && minutes! >= 0 && minutes! <= 59 && // here the example with the single digit would pass the check
        Number.isInteger(seconds) && seconds! >= 0 && seconds! <= 59
    );
```

### new function
- *the new checkTimeFormat() looks like this in code: `const checkTimeFormat = (time: string) => { return regTime(time); };`*
	- below is the relevant code from `reggex.ts`

```typescript
const timeReg: RegExp = /^([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])$/

export function regTime(timestamp: string): boolean {
    return timeReg.test(timestamp);
}
```

## Used wrong default value for no leader selection (on creation) (FE)
- the leader input in the form has the default value of an empty string: `<option value="">Select leader</option>` (line 377)
	- this empty string value would be loaded into the request, **when the user was not a leader, but a BOARD_MEMBER or ADMIN**, which then would fail the Zod parsing in the backend instead of defaulting to the current user
	- solution: added the missing roles to the isLeader assignement:
		- `const isLeader = user?.role === (ROLES.LEADER || ROLES.BOARD_MEMBER || ROLES.ADMIN);`

## Failed to load existing leaders into ActivityForm (on edit) (FE)
- when editing an activity, one had to select an activity leader in the form, even if leaders already exist.
*Solution*
```typescript
const leaderIds = json.data.map(el => el.id);
setSelectedLeaderIds(leaderIds);
```
**Note, I think this just loads all the leaders into the form, I think we should probably add a filter method here that filters out the current activities leaders and loads them into the form, something like this:**
```typescript
const leaderIds = json.data.filter(id => id in activity.leaders);
```

---

# 409-repsonse for POST newActivity
- queries DB for name of the new activity
    - if something is returned, a 409-conflict error is sent back
    - else the new activity is created

---

# Files Touched/Created:

### activities.controller.ts
- added Zod-parsing error handling (apparently didn't have it)

### createQueries.ts
- added check for existing activity (by activity name)
	- if an activity exists, a 409-conflict error is thrown
	- else the activity is successfully created

### updateQueries
- added `await this.deleteAllTimeSlots(activityId)` because frontend sends also the existing timeSlots with the update, would otherwise throw Unique Constraint violations.
### errorHandler
- added if-statement as instance check of `PrismaClientKnownRequestError`
```typescript
if( err instanceof PrismaClientKnownRequestError ){
	const prismaErr = prismaErrorFormatter(err);
	res.status(400).json({
	  error: prismaErr,
	  statusCode: 400,
	})
}
```

### reggex.ts
- adjusted the time reggex to reflect current format of `HH:mm:ss`
	- this had to be done due to the `checkTimeFormat()` function not covering all cases. Look at the **Bugs** heading.

### activity.validator.ts
- replaced function-body of `checkTimeFormat()` due to past implementation not catching all errors.  Look at the **Bugs** heading.

### (New) prismaErrorFormatter.ts
- formats any given `PrismaClientKnownRequestError` to only contain crucial information
```typescript
import {PrismaClientKnownRequestError} from "@prisma/client/runtime/client";

export function prismaErrorFormatter(error: PrismaClientKnownRequestError) {
    const driverAdaptErr = error.meta?.driverAdapterError as any;
    const simpleErr =  {
        code: error.code,
        name: driverAdaptErr.cause.kind,
        fields: driverAdaptErr.cause.constraint.fields,
        table: error.meta?.modelName,
        message: driverAdaptErr.cause.originalMessage
    }
    return simpleErr;
}
```

## Frontend (ActivityFormPage.jsx)
- fixed a bug, where not selecting a leader would default to the value given in the form (an empty string), which would fail the UUID parsing in the backend, fix is below:
	- added Board_Member and Admin roles to the `isLeader` statement (line 30)
		- `const isLeader = user?.role === (ROLES.LEADER || ROLES.BOARD_MEMBER || ROLES.ADMIN);`
- For editing an activity, the page would not load the existing leaders into the form, which would cause another error return, as at least one leader is required. Added the lines below into the fetching functions of leaders (line 115 & 116):
	- this displays the first leader in the array in the form and loads the entire array into the request object
```typescript
const leaderIds = json.data.map(el => el.id);
setSelectedLeaderIds(leaderIds);
```


JuriRappold:
For how the 400-response looks like for prisma client errors:
```json
{
  "error": {
    "code": "P2002",
    "name": "UniqueConstraintViolation",
    "fields": [
      "\"activityId\"",
      "weekday",
      "\"startTime\""
    ],
    "table": "ActivityTemplate",
    "message": "duplicate key value violates unique constraint \"TimeSlot_activityId_weekday_startTime_key\""
  },
  "statusCode": 400
}
```
Might be too much information revealed to the client, but thats an easy fix.

client/src/pages/ActivityFormPage.jsx
```diff
@@ -27,7 +27,7 @@ export default function ActivityFormPage() {
   const navigate = useNavigate()
   const { id: activityId } = useParams()
   const isEditMode = Boolean(activityId)
-  const isLeader = user?.role === ROLES.LEADER
+  const isLeader = user?.role === (ROLES.LEADER || ROLES.BOARD_MEMBER || ROLES.ADMIN);
```
gunnarwrld: The `||` operator returns the first truthy value, so `ROLES.BOARD_MEMBER` and `ROLES.ADMIN` are never evaluated. Fix:
```
const isLeader = [ROLES.LEADER, ROLES.BOARD_MEMBER, ROLES.ADMIN].includes(user?.role);
```

server/src/utils/prismaErrorFormatter.ts
```diff
@@ -0,0 +1,13 @@
+import {PrismaClientKnownRequestError} from "@prisma/client/runtime/client";
+
+export function prismaErrorFormatter(error: PrismaClientKnownRequestError) {
+    const driverAdaptErr = error.meta?.driverAdapterError as any;
```
gunnarwrld: `driverAdapterError` is only populated when using a driver adapter (like `@prisma/adapter-pg`). For other `PrismaClientKnownRequestError` subtypes (e.g., P2025 record-not-found, P2003 FK violation), `meta.driverAdapterError` is `undefined`, causing an unhandled `TypeError` that crashes the error handler itself. The server then falls through to the generic 500 branch (worse than before lmao). Add null guards:
```
export function prismaErrorFormatter(error: PrismaClientKnownRequestError) {
    const driverAdaptErr = error.meta?.driverAdapterError as any;
    return {
        code: error.code,
        name: driverAdaptErr?.cause?.kind ?? error.code,
        fields: driverAdaptErr?.cause?.constraint?.fields ?? error.meta?.field_name,
        table: error.meta?.modelName,
        message: driverAdaptErr?.cause?.originalMessage ?? error.message,
    };
}
```

server/src/middleware/errorHandler.ts
```diff
@@ -18,6 +20,13 @@ export function errorHandler(
     });
     return;
   }
+  if( err instanceof PrismaClientKnownRequestError ){
+    const prismaErr = prismaErrorFormatter(err);
+    res.status(400).json({
```
gunnarwrld: After `res.json()` the function continues executing the `console.error('Unhandled error:', err)` and then tries to call `res.status(500)` on an already-sent response, throwing a "Cannot set headers after they are sent" error. Add return.

server/src/middleware/errorHandler.ts
```diff
@@ -18,6 +20,13 @@ export function errorHandler(
...
+      error: prismaErr,
+      statusCode: 400,
```
gunnarwrld: Not all known Prisma errors are client faults. P2001/P2025 (record not found) are 404, P2002 (unique constraint) is 409. Hardcoding 400 for every `PrismaClientKnownRequestError` is not correct and misleads callers

server/src/middleware/errorHandler.ts
```diff
@@ -1,5 +1,7 @@
 import type { Request, Response, NextFunction } from 'express';
 import { ApiError } from '../utils/ApiError.js';
+import {PrismaClientKnownRequestError} from "@prisma/client/runtime/client";
```
gunnarwrld: `@prisma/client/runtime/client` is an internal runtime path, not part of Prisma's public API; it can silently break on any minor Prisma upgrade. Use the re-exported public symbol instead:
```
import { Prisma } from '../generated/prisma/index.js';
// then:
if (err instanceof Prisma.PrismaClientKnownRequestError)
```

server/src/utils/prismaErrorFormatter.ts
```diff
@@ -0,0 +1,13 @@
+import {PrismaClientKnownRequestError} from "@prisma/client/runtime/client";
```
gunnarwrld: Same import issue here

server/src/db/createQueries.ts
```diff
@@ -44,43 +44,54 @@ export class CREATE {
...
-                        }))
-                    }
-                },
-            },
-            include: {
-                timeSlots: true,
-                leaders: {
-                    select: {
-                        profile: {
+        const activity = await prisma.$transaction(async (tx) => {
```
gunnarwrld: Two concurrent requests can both pass the `findFirst` check and both attempt create. If the DB has a unique constraint on `activityTemplate.name`, the second will throw a constraint error (handled by `prismaErrorFormatter`). If there is no unique constraint, both silently succeed. We have to make sure that the schema has a `@@unique([name])` constraint, and rely on that as the source of truth; the application-level check is redundant and still racy.

server/src/db/updateQueries.ts
```diff
@@ -16,21 +16,20 @@ export class UPDATE {
...
         await prisma.activityTemplate.update({
-            where: { id: activityId },
+            where: {id: activityId},
             // @ts-ignore - this is a bit hacky, but it allows us to only include fields that are actually being updated (excludes timeSlots and leaders)
             data: newData,
         });
-
         if (!!timeSlots) {
+            await this.deleteAllTimeSlots(activityId);
             await this.addTimeSlots(activityId, timeSlots);
```
gunnarwrld: If `addTimeSlots` throws, the activity is left with no time slots. Wrap both steps (and the leader replacement below) in a `prisma.$transaction`.

client/src/pages/ActivityFormPage.jsx
```diff
@@ -111,8 +111,9 @@ export default function ActivityFormPage() {
         if (!json) return
 
         if (json.status === 'success') {
-          //console.log(json.data)
           setLeaders(json.data)
+          const leaderIds = json.data.map(el => el.id);
+          setSelectedLeaderIds(leaderIds);
```
gunnarwrld: The `/leaders` endpoint returns all available leaders in the system, not the current activity's leaders. This pre-selects every leader in the activity on the edit form.Also, `id in activity.leaders` uses `in` for object keys, not array membership. Use the activity's own leader list from the activity fetch:
```
const currentLeaderIds = activity.leaders.map(l => l.id);
setSelectedLeaderIds(currentLeaderIds);
```

gunnarwrld (changes requested):
Great work overall Juri, the PR addresses real pain points (unhandled Prisma errors, the time slot regex gap, and the 409 conflict response) and the transaction wrapping in `newActivity` is the right direction. A few things need to be fixed before this can merge.

jasminerezai (changes requested):
The intent is right but there are multiple blocking issues: the error handler crashes on the most common Prisma error types (making things worse than before), the isLeader fix doesn't actually work, and the edit-form leader pre-selection bug is a data-integrity issue. Address @gunnarwrld comments directly and push a follow-up commit. You are close close, just needs the fixes applied.

client/src/pages/ActivityFormPage.jsx
JuriRappold: I for rn, added the the loading of the leaders into the fetching of the activity, making the fetch call for the leaders redundant if the page is in edit mode, i think
And commented out the the above lines

JuriRappold:
# Implemented better error handling
- switch-case statement for the prismErrorFormatter function
- return statement added to global errorHandler

# $transaction wrapper
- delete & create statements of timeSlots and leaders were wrapped in a `$transaction` block

# unique constraint in ActivityTemplate
- the column name is now unique
	- initially also added the unique constraint to the Schedule, but that would have also required handling in the cron-job. Is also its *own* issue #114
	- removed the schedule constraint

# Bug fixes in ActivityFormPage.jsx
- implemented @gunnarwrld  correct if-statement for `isLeader`
	- this revealed buggy conditional rendering
		- solution: inverting the condition for some rendering to `!isLeader`

client/src/pages/ActivityFormPage.jsx
```diff
@@ -27,7 +27,7 @@ export default function ActivityFormPage() {
   const navigate = useNavigate()
   const { id: activityId } = useParams()
   const isEditMode = Boolean(activityId)
-  const isLeader = user?.role === ROLES.LEADER
+  const isLeader = [ROLES.LEADER, ROLES.BOARD_MEMBER, ROLES.ADMIN].includes(user?.role);
```
gunnarwrld: Leaders/admins now get the restricted body and disabled fields; non-privileged users get the full form. Every `!isLeader` in this file must be reverted to `isLeader`. The widened definition on this line is correct, all the negations below are not.

client/src/pages/ActivityFormPage.jsx
```diff
@@ -158,8 +160,7 @@ export default function ActivityFormPage() {
     }
 
     setLoading(true)
-
-    const body = isLeader
+    const body = !isLeader
```
gunnarwrld: Direct consequence of the `isLeader` inversion above. `!isLeader` here means an admin/board member submits the restricted body (location + description only), while a regular user submits the full body including `leaders` and `timeSlots`. Revert to `isLeader`.

server/src/db/updateQueries.ts
```diff
@@ -16,31 +16,35 @@ export class UPDATE {
...
             data: newData,
         });
-
         if (!!timeSlots) {
-            await this.addTimeSlots(activityId, timeSlots);
+            await prisma.$transaction( async () => {
+                // does this function call work in the $transaction wrapper?
+                await this.deleteAllTimeSlots(activityId);
+                await this.addTimeSlots(activityId, timeSlots);
+            })
```
gunnarwrld: The general field update (`activityTemplate.update`) runs outside any transaction. If the subsequent time slot or leader transactions fail, the activity is left partially updated. Wrap the entire `updateActivity` body in a single `prisma.$transaction(async (tx) => { ... })` and pass `tx` to all inner operations.

server/src/db/updateQueries.ts
```diff
@@ -16,31 +16,35 @@ export class UPDATE {
...
-            });
-            await prisma.leaderActivity.createMany({
-                data: leaders.map(profileId => ({
-                    profileId,
-                    activityId
-                }))
-            });
+            await prisma.$transaction(async(tx) => {
+                await tx.leaderActivity.deleteMany({
+                    where: {activityId}
```
gunnarwrld, line 30: You already flagged this with the comment `// does this function call work in the $transaction wrapper?`, it does not. `deleteAllTimeSlots` and `addTimeSlots` use the outer `prisma` instance, not the transaction client `tx`. Calls made through `prisma` inside a `$transaction` callback run on a separate connection and are not part of the transaction. The atomicity guarantee is lost.

Refactor the helpers to accept a `tx` parameter, then pass the transaction client through:
```ts
await prisma.$transaction(async (tx) => {
    await this.deleteAllTimeSlots(activityId, tx);
    await this.addTimeSlots(activityId, timeSlots, tx);
});
```

server/src/db/createQueries.ts
```diff
@@ -44,43 +44,54 @@ export class CREATE {
...
+            const activityExists = await tx.activityTemplate.findFirst({
+                where: {name: newAct.name}
+            });
+            if (activityExists) {
+                const format = formatActivity(activityExists)
```
gunnarwrld: The `findFirst` + `throw ApiError.conflict` pre-check is redundant now that `@@unique([name])` is enforced at the DB level via the migration in this PR. This adds an extra roundtrip and a TOCTOU race window between the `findFirst` and the `create`. Remove the manual check and let the unique constraint fire a Prisma `P2002` error, which `errorHandler.ts` already catches and maps to 409 via `prismaErrorFormatter`.

server/src/utils/prismaErrorFormatter.ts
```diff
@@ -0,0 +1,94 @@
+import { Prisma } from '../generated/prisma/index.js';
+
+export function prismaErrorFormatter(error: Prisma.PrismaClientKnownRequestError): {error: object, statusCode: number} {
+    const driverAdaptErr = error.meta?.driverAdapterError as any;
+    const nicerMsg = error.message.split('\n');
```
gunnarwrld, line 4: The return type `{error: object, statusCode: number}` breaks the codebase error contract, which is `{ error: string, statusCode: number }` per `ApiError` and `errorHandler`. `simpleErr` here is an object, not a string. Flatten it to a string or align the contract explicitly:
```ts
export function prismaErrorFormatter(error: Prisma.PrismaClientKnownRequestError): { error: string; statusCode: number }
```
and then `error: simpleErr.message` (or a composite string).

server/src/utils/prismaErrorFormatter.ts
```diff
@@ -0,0 +1,94 @@
...
+        name: driverAdaptErr?.cause.kind ?? error.code,
+        // fields: driverAdaptErr?.cause?.constraint?.fields ?? error.meta?.field_name,
+        message: nicerMsg[nicerMsg.length - 1] //driverAdaptErr?.cause?.originalMessage ??
+    }
+    const errObject = {
+        error: simpleErr,
+        statusCode: 400
+    }
+    const codeAsNum: number = Number(simpleErr.code.slice(1));
+    // Reference: https://www.prisma.io/docs/orm/reference/error-reference
```
gunnarwrld: `switch(true)` with equality cases is a disguised `if/else if` chain and harder to read. Replace with this alternative of a lookup map i found:
```ts
const statusMap: Record<number, number> = {
  2001: 404, 2002: 409, 2003: 409, 2014: 409,
  2015: 404, 2018: 404, 2021: 404, 2022: 404,
  2024: 500, 2025: 404, 2026: 500, 2027: 500, 2028: 500, 2030: 404,
};
errObject.statusCode = statusMap[codeAsNum] ?? (codeAsNum > 2030 ? 500 : 400);
```
Also note: this file belongs in `middleware/` not `utils/`, it is tightly coupled to the Express error response shape and is only called from `errorHandler.ts`.

server/prisma/migrations/20260526150408_removed_constraint_from_schedule/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- DropIndex
+DROP INDEX "Schedule_activityId_startAt_key";
```
gunnarwrld: This migration immediately drops the index created in `20260526144445_unique_constraint_name_and_schedule`. Two consecutive create-then-drop migrations for the same constraint should be squashed into one migration containing only the final desired state (i.e., just the `ActivityTemplate_name_key` unique index). Having this pair in migration history is misleading and adds a redundant step to every fresh `migrate deploy`.

server/src/utils/reggex.ts
```diff
@@ -8,8 +8,9 @@
...
  */
-const timeReg: RegExp = /^1970-01-01T([01][0-9]|2[0-3]):([0-5][0-9]):00\+01:00$/
+const timeReg: RegExp = /^([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])$/
 
 export function regTime(timestamp: string): boolean {
     return timeReg.test(timestamp);
-}
\ No newline at end of file
+}
+
```
gunnarwrld: Trailing whitespace on this line after the closing `}`. Remove it. Also the file still ends without a newline (`\ No newline at end of file`), add one.

gunnarwrld (changes requested):
Great work Juri, the Prisma error formatter is a solid addition and the favorites feature is clean. Just a couple of blockers before merge!

JuriRappold:
## Changes made:
1. reverted the `!isLeader` in `ActivityFormPage.jsx`
2. *updateQueries.ts*: wrapped all queries in `$transaction()`
	1. replaced calls to deleteAllTimeSlots & addTimeSlots with copied query (can't call an external function in $transaction if we want to properly use $transaction)
	2. saved the return values of the queries to return the updated object, made the last query redundant
3. *createQueries.ts*: removed redundant existence check
4. *prisma/migrations/* removed last migration & removed creation of a restraint in second-to-last
	1. removed constraint is being added in PR: #119 or #122
		1. the `@@unique([activityId, startAt])` constraint
5. *prismaErrorFormatter*:
	1. corrected the error shape to a string
		1. From: `error: simpleErr`
		2. To: `error: '${simpleErr.name} : ${simpleErr.message}'`
	2. replaced switch-case statement with map
	3. moved file to `middleware/`

client/src/pages/ActivityFormPage.jsx
larrymijo: Broadening `isLeader` to include BOARD_MEMBER and ADMIN inverts the access logic for those roles. `isLeader` is used throughout this file to mean "restricted role, lock things down": fields are gated with `disabled={loading || isLeader}` (name, max capacity, leader select, time slot inputs, add/remove slot) and helper texts render with `{isLeader && (...)}` saying "This field can only be edited by admins or board members." With this change, board members and admins now get those fields disabled and shown a message telling them they can't edit.

More seriously, `handleSubmit` does `const body = isLeader ? {restricted} : {full}`, and the restricted branch omits `name`, `maxCapacity`, `leaders`, and `timeSlots`. So board members/admins editing an activity will now silently strip those fields on save — a data-integrity regression.

The fix should introduce a separate concept rather than overloading `isLeader`, e.g. a dedicated `isRestrictedLeader` (only `=== ROLES.LEADER`) for the disable/body logic, and a `canEditAll`/`canManage` flag for the privileged roles. Right now the single variable is being used for two contradictory purposes.

client/src/pages/ActivityFormPage.jsx
```diff
@@ -111,8 +112,9 @@ export default function ActivityFormPage() {
         if (!json) return
 
         if (json.status === 'success') {
-          //console.log(json.data)
           setLeaders(json.data)
+          // const leaderIds = json.data.map(el => el.id);
```
larrymijo: Nit: these two commented-out lines are dead code (the real pre-selection now happens in the edit-mode effect via `setSelectedLeaderIds(a.leaders.map(l => l.id))`). Please remove them rather than leaving them commented out.

larrymijo (changes requested):
Reviewed focusing on the front end (ActivityFormPage.jsx). The intent here is good and the edit-mode leader pre-selection fix is a real improvement, but there are a couple of blocking issues before this can merge.

1. The `isLeader` change on line 30 is currently a regression for board members and admins — see my inline comment. Because `isLeader` is reused both to disable fields/show "only admins or board members can edit" helper text AND to choose the restricted submit body (which drops name, maxCapacity, leaders, and timeSlots), broadening it to include BOARD_MEMBER/ADMIN actually locks those roles out and silently strips fields on save. This needs to be split into two separate flags.
2. Separate from this diff, the leader `<select>` (around line 363) only supports a single leader: `value={selectedLeaderIds[0]}` plus `onChange={(e) => setSelectedLeaderIds([e.target.value])}` collapses the array to one ID. The new edit-mode fix correctly loads all existing leaders, but the moment the user touches this control every leader except one is discarded on save. Worth addressing (multi-select or chips) since it interacts directly with the leader-loading fix in this PR.
Also left a small nit about removing the commented-out dead code at lines 116-117. Happy to re-review once these are addressed — thanks!

client/src/pages/ActivityFormPage.jsx
JuriRappold: ahh Ok I think I misunderstood the boolean in this case: I thought it granted access, but instead it restricts access. I'll revert this change then

client/src/pages/ActivityFormPage.jsx
```diff
@@ -360,8 +359,15 @@ export default function ActivityFormPage() {
           </label>
 
           <select
-            value={selectedLeaderIds[0] || ''}
-            onChange={(e) => setSelectedLeaderIds([e.target.value])}
+            value={selectedLeaderIds || ''}
```
gunnarwrld: `value={selectedLeaderIds || ''}` passes an array to a `<select>` value prop. A `<select>` is not a multi-select here (no `multiple` attribute), so setting its value to an array is undefined behavior in React, it will render as `[object Object]` or the first element depending on the React version. The original `value={selectedLeaderIds[0] || ''}` was correct for a single-select. If the intent is to allow multiple leaders, add `multiple` to the element and adjust the label accordingly. If it is still single-select, revert to `selectedLeaderIds[0] || ''`.

client/src/pages/ActivityFormPage.jsx
```diff
@@ -360,8 +359,15 @@ export default function ActivityFormPage() {
...
+            onChange={(e) => {
+              let set = new Set(selectedLeaderIds.map( el => el));
+              set.add(e.target.value);
```
gunnarwrld: Unnecessary manual `Set` + `forEach` roundtrip. `new Set(selectedLeaderIds.map(el => el))` is also identity-mapping the array for no reason (`new Set(selectedLeaderIds)` is equivalent). And the whole block can be replaced with:

```js
onChange={(e) => {
  const next = [...new Set([...selectedLeaderIds, e.target.value])];
  setSelectedLeaderIds(next);
}}
```

Also note: this `onChange` only ever adds leaders, never removes them. Once a leader is added via the dropdown, there is no way to deselect them.

server/src/middleware/prismaErrorFormatter.ts
```diff
@@ -0,0 +1,25 @@
+import { Prisma } from '../generated/prisma/index.js'
+
+export function prismaErrorFormatter(error: Prisma.PrismaClientKnownRequestError): {error: string, statusCode: number} {
+    const driverAdaptErr = error.meta?.driverAdapterError as any;
+    const nicerMsg = error.message.split('\n');
+    const simpleErr =  {
+        code: error.code,
+        name: driverAdaptErr?.cause.kind ?? error.code,
```
gunnarwrld: `driverAdaptErr?.cause.kind`, the optional chain stops at `driverAdaptErr` but then accesses `.cause.kind` without guarding. If `driverAdaptErr` is defined but `cause` is undefined (e.g. a Prisma error that does not originate from the pg driver adapter), this throws `TypeError: Cannot read properties of undefined (reading 'kind')` and crashes the error handler, turning a structured error response into an unhandled crash.

Guard the entire chain:
```ts
const name = driverAdaptErr?.cause?.kind ?? error.code;
const message = nicerMsg[nicerMsg.length - 1] ?? error.message;
```

server/src/db/updateQueries.ts
```diff
@@ -1,61 +1,77 @@
 import { prisma, ActivityStatus } from "./prisma.js";
-import {Activity, TimeSlot, ScheduleDto, ActivityDto} from "../types/index.js";
+import {Activity, ScheduleDto, ActivityDto, leaderDto} from "../types/index.js";
 import {formatActivity, formatSchedule} from "./utils.js";
 import {ApiError} from "../utils/ApiError.js";
+import { TimeSlot } from '../generated/prisma/index.js';
```
gunnarwrld: I found the problem here Juri, `TimeSlot` is imported from `'../generated/prisma/index.js'` (the Prisma-generated model type), but the `Activity` type from `'../types/index.js'` has its own `TimeSlot` shape defined in `activity.types.ts`. These are two different types with different field names, the Prisma `TimeSlot` has `startTime`/`endTime`, the domain `TimeSlot` has `startAt`/`endAt`. 

Importing the Prisma type here is what drove the field rename in `addTimeSlots` from `el.startAt` to `el.startTime`. This is a convention violation: all types should come from the `'../types'` barrel, never from individual domain files or the generated client directly. Align the import and the domain type shape instead.

server/src/db/updateQueries.ts
```diff
@@ -1,61 +1,77 @@
...
+            if (!!slots) {
+                await tx.timeSlot.deleteMany({
+                    where: {activityId}
+                });
 
-        if (!!timeSlots) {
-            await this.addTimeSlots(activityId, timeSlots);
-        }
+                newTimes =  await tx.timeSlot.createManyAndReturn({
+                    data: slots.map(el => ({
```
gunnarwrld, line 34: The `satisfies ActivityDto` constraint on the transaction return object is misleading. `ActivityDto` has a `timeSlots` field typed as the domain `TimeSlot[]` (with `startAt`/`endAt`), but `newTimes` is `TimeSlot[]` from the Prisma model (with `startTime`/`endTime`). TypeScript accepts the `satisfies` here only because the import mismatch above masks the incompatibility (at runtime), `formatActivity(newAct)` will receive Prisma `TimeSlot` objects in the `timeSlots` field and may map them incorrectly. Verify what `formatActivity` expects and ensure the types align end-to-end.

client/src/pages/ActivityFormPage.jsx
```diff
@@ -178,7 +177,7 @@ export default function ActivityFormPage() {
           leaders: selectedLeaderIds,
           timeSlots,
         }
-
+    console.log(body);
```
gunnarwrld: `console.log(body)` left in a production path. This logs the full activity body including `leaders`, `timeSlots`, and all form fields on every submit. Remove it before merging.

server/src/db/updateQueries.ts
```diff
@@ -1,61 +1,77 @@
...
+
+            return {
+                ...general,
+                leaders: newLeaders,
+                timeSlots: newTimes
+            } satisfies ActivityDto;
+        })
+
+        // do we still need this check?
+        if (!newAct) throw ApiError.notFound(`Activity to update not Found`);
```
gunnarwrld: No, `if (!newAct) throw ApiError.notFound(...)` is now unreachable. `prisma.$transaction` either returns the constructed object or throws, it never returns `null` or `undefined`. Remove it

server/src/utils/prismaErrorFormatter.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+    const codeAsNum: number = Number(simpleErr.code.slice(1));
+
+    // Reference: https://www.prisma.io/docs/orm/reference/error-reference
+    const statusMap: Record<number, number> = {
+        2001: 404, 2002: 409, 2003: 409, 2014: 409,
+        2015: 404, 2016: 404, 2018: 404, 2021: 404, 2022: 404,
+        2024: 500, 2025: 404, 2026: 500, 2027: 500, 2028: 500, 2030: 404,
+    };
+    errObject.statusCode = statusMap[codeAsNum] ?? (codeAsNum > 2030 ? 500 : 400);
```
gunnarwrld, line 21: The status code mapping has a bug: Prisma error code `P2002` (unique constraint violation, the very case this PR introduces for `ActivityTemplate.name`) maps to `codeAsNum = 2002`, which is mapped to `409`. That is correct. But `P2025` (record not found, used when `deleteActivity` fails) maps to `404`, also correct. However `P2003` (foreign key constraint failure) maps to `409 Conflict` which is semantically wrong; a foreign key violation should be `400 Bad Request` or `422 Unprocessable Entity`, not `409`. `409` implies a state conflict with an existing resource, not a referential integrity failure.

Also, the `deleteActivity` controller was changed in this PR to remove the pre-existence check and rely on Prisma throwing `P2025`. Verify the error handler now returns a proper 404 JSON body in the `{ error: string, statusCode: number }` contract shape for that case.

gunnarwrld (changes requested):
Good direction overall. Wrapping create/update in $transaction, centralizing Prisma error handling in the middleware, and fixing the regex validator are all correct. Eight issues, two of which are blockers before merging!

server/src/db/updateQueries.ts
JuriRappold: Issue is that `ActivityDto` does not use the TimeSlot type from `activity.types.ts`.  Also `formatActivity` just assigns the timeSlots like this: `timeSlots: activity.timeSlots`

client/src/pages/ActivityFormPage.jsx
```diff
@@ -178,7 +177,7 @@ export default function ActivityFormPage() {
           leaders: selectedLeaderIds,
           timeSlots,
         }
```
gunnarwrld, line 179: still this console.log(body) left in production path. This logs the full activity payload on every create/update submit (potential data leakage and a direct violation of the enforced conventions). Remove it before merge.
-    console.log(body);

client/src/pages/ActivityFormPage.jsx
```diff
@@ -27,7 +27,7 @@ export default function ActivityFormPage() {
   const navigate = useNavigate()
   const { id: activityId } = useParams()
   const isEditMode = Boolean(activityId)
-  const isLeader = user?.role === ROLES.LEADER
+  const isLeader = user?.role === ROLES.LEADER;
```
gunnarwrld, line 30: **`isLeader` fix is broken -- `||` on strings does not work the way you expect.** `ROLES.LEADER || ROLES.BOARD_MEMBER || ROLES.ADMIN` short-circuits to `ROLES.LEADER` (the first truthy string), so `user?.role === (ROLES.LEADER || ROLES.BOARD_MEMBER || ROLES.ADMIN)` is identical to the old single-role check. The bug described in the PR description is NOT fixed by this line.

Fix with an explicit inclusion check:
```js
const isLeader = [ROLES.LEADER, ROLES.BOARD_MEMBER, ROLES.ADMIN].includes(user?.role);
```

Also note: the actual diff only changes the semicolon on this line. The role expansion described in the PR body does not appear in the diff at all, it is either missing from this branch or was accidentally omitted (by lawrence).

client/src/pages/ActivityFormPage.jsx
```diff
@@ -88,6 +88,7 @@ export default function ActivityFormPage() {
               endAt:   new Date(slot.endTime).toISOString().slice(11, 19),
             })))
           }
+          setSelectedLeaderIds(a.leaders.map( l => l.id));
```
gunnarwrld, line 91: **`setSelectedLeaderIds` called with raw leader IDs from the activity fetch, not from the `/leaders` endpoint response.** `a.leaders.map(l => l.id)` is correct, it seeds the edit form with existing leaders. This is the right fix. One edge case to be aware of: if `a.leaders` is undefined or empty (e.g. activity was saved without leaders), this will call `setSelectedLeaderIds([])` which is safe, so no issue there.

server/src/utils/prismaErrorFormatter.ts
```diff
@@ -0,0 +1,25 @@
+import { Prisma } from '../generated/prisma/index.js'
+
+export function prismaErrorFormatter(error: Prisma.PrismaClientKnownRequestError): {error: string, statusCode: number} {
+    const driverAdaptErr = error.meta?.driverAdapterError as any;
+    const nicerMsg = error.message.split('\n');
```
gunnarwrld, line 4: **`driverAdapterError` access is unsafe and will throw on most Prisma errors.** `error.meta?.driverAdapterError as any` only exists when using the `@prisma/adapter-pg` driver adapter, and even then only for specific errors. For the majority of `PrismaClientKnownRequestError` cases (e.g. P2002 unique constraint, P2025 record not found) `driverAdapterError` will be `undefined`, so `driverAdaptErr?.cause?.kind` silently falls back to `error.code`, which is fine. But `driverAdaptErr.cause.originalMessage` on line 12 (in the PR body's version) would throw. The actual diff uses `nicerMsg[nicerMsg.length - 1] ?? error.message` which is safer.

The PR description's code snippets are out of sync with what is actually in the diff. The merged implementation is the correct one.

server/src/middleware/prismaErrorFormatter.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    // Reference: https://www.prisma.io/docs/orm/reference/error-reference
+    const statusMap: Record<number, number> = {
+        2001: 404, 2002: 409, 2003: 400, 2014: 409,
+        2015: 404, 2016: 404, 2018: 404, 2021: 404, 2022: 404,
+        2024: 500, 2025: 404, 2026: 500, 2027: 500, 2028: 500, 2030: 404,
+    };
+    errObject.statusCode = statusMap[codeAsNum] ?? (codeAsNum > 2030 ? 500 : 400);
+    return errObject;
+}
```
gunnarwrld: **Missing newline at end of file.** Minor, but will cause a linter/diff noise on the next PR that touches this file. Add a trailing newline.

server/src/middleware/errorHandler.ts
```diff
@@ -18,6 +20,11 @@ export function errorHandler(
     });
     return;
   }
+  if( err instanceof Prisma.PrismaClientKnownRequestError ){
+    const prismaErr = prismaErrorFormatter(err);
```
gunnarwrld, line 24: **`prismaErrorFormatter` is imported from `./prismaErrorFormatter.js` (a sibling middleware file), not from the `utils/` directory.** Utility functions that are not Express middleware should live in `server/src/utils/`, not `server/src/middleware/`. The formatter has no dependency on Express types, it takes a Prisma error and returns a plain object. Move `prismaErrorFormatter.ts` to `server/src/utils/prismaErrorFormatter.ts` and update the import here.

server/src/db/updateQueries.ts
```diff
@@ -1,62 +1,73 @@
...
-                        profile: {
-                            select: {
-                                id: true,
-                                profileName: true
+            if (!!leaders) {
+                    await tx.leaderActivity.deleteMany({
+                        where: {activityId}
+                    });
+                    newLeaders = await tx.leaderActivity.createManyAndReturn({
+                        data: leaders.map(profileId => ({
```
gunnarwrld, line 54: **Good: `updateActivity` is now fully inside a `prisma.$transaction`.** The old implementation did general-field update, then time slot delete/create, then leader delete/create as three separate operations, a crash between any two would leave the DB in a partial state. Wrapping all of this in one transaction is the correct fix.

server/src/db/updateQueries.ts
```diff
@@ -1,62 +1,73 @@
...
+                ...general,
+                leaders: newLeaders,
+                timeSlots: newTimes
+            }
+        })
+
+        return formatActivity(newAct);
     }
 
     /**
```
gunnarwrld, line 92: **`formatActivity` receives a synthetic object, not a real Prisma query result (type mismatch risk).** The transaction returns `{ ...general, leaders: newLeaders, timeSlots: newTimes }` where `newLeaders` is `[]` when no leader update happened, and `newTimes` is `[]` when no slot update happened. If `formatActivity` iterates over `leaders` or `timeSlots` and expects the full nested shape from a `findUnique` include, it will silently return an activity with empty leaders/timeSlots even though the DB still has the old ones.

Either (a) do a final `tx.activityTemplate.findUnique(...)` at the end of the transaction and pass that to `formatActivity`, or (b) ensure `formatActivity` handles the partial shape and that callers understand the returned DTO may not reflect unchanged relations.

server/src/db/createQueries.ts
```diff
@@ -44,44 +44,46 @@ export class CREATE {
...
-                        }))
-                    }
-                },
-            },
-            include: {
-                timeSlots: true,
-                leaders: {
-                    select: {
-                        profile: {
+        const activity = await prisma.$transaction(async (tx) => {
```
gunnarwrld: **`newActivity` wrapping in `prisma.$transaction` is good, but the transaction adds no real value here.** A single `prisma.activityTemplate.create` with nested `createMany` is already atomic in Prisma; the nested writes are issued inside Prisma's own implicit transaction. Wrapping it in an explicit `$transaction` with only one statement is harmless but misleading, it suggests there are multiple operations that need to be coordinated. If the 409-conflict check (querying by name before create) described in the PR body is intended to be inside this transaction to prevent races, it is not in the diff. If that check is the reason for the transaction, add it here; otherwise remove the wrapping and keep the flat create.

server/src/controllers/activities.controller.ts
```diff
@@ -143,7 +145,8 @@ export const newActivity = asyncHandler(
     try {
       newActivity = CreateActivitySchema.parse(req.body);
     } catch (error) {
```
gunnarwrld, line 147: **Good: Zod parse error now distinguishes `ZodError` from unknown errors.** The old implementation stringified any error as a bad request. The new split -- `ZodError` -> 400 with parsed field errors, everything else -> 500 -- is correct. `parseZodError` should format field-level messages; make sure it is already exported from `'../types'` or `'../utils'` and not from an individual file.

gunnarwrld (changes requested):
Good effort overall, the transaction refactor in `updateActivity`, the `prisma.$transaction` in `newActivity`, the `@@unique` on `ActivityTemplate.name`, and the `PrismaClientKnownRequestError` handling in `errorHandler` are all solid additions that address real gaps. The regex fix for time validation is also correct and cleaner than the old manual split.

That said, there are a few things that need to be addressed before this merges:

server/src/utils/prismaErrorFormatter.ts
JuriRappold: I think ill just remove the driverAdapter stuff, Don't think we need such a specific error message for the frontend

server/src/middleware/errorHandler.ts
JuriRappold: > Also note: this file belongs in `middleware/` not `utils/`, it is tightly coupled to the Express error response shape and is only called from `errorHandler.ts`.

Mate, I totally agree, but in an earlier change request you requested it to be moved to `middleware/`. Will move it back

server/src/db/updateQueries.ts
JuriRappold: Good Catch! Resolved with these two if-statements:
```typescript
if(!leaders){
	newLeaders = await tx.leaderActivity.findMany({
		where: {activityId},
		select: {
			profile: {
				select: {
					id: true,
					profileName: true
				}
			}
		}
	})
}
if(!slots){
	newTimes = await tx.timeSlot.findMany({
		where: {activityId},
	})
}
```

client/src/pages/ActivityFormPage.jsx
JuriRappold: Hmm no I removed it and reverted it to its original shape. I misunderstood the boolean function, I thought it granted access to the certain fields, but the opposite is true. It restricts access. The intial bug was solved by correctly loading the default value into the request body, therefore this:
```typescript
const isLeader = [ROLES.LEADER, ROLES.BOARD_MEMBER, ROLES.ADMIN].includes(user?.role);
```
is no longer neccessary.

gunnarwrld:
The real blockers from earlier rounds have all been addressed. Feel free to merge!!

gunnarwrld (approved):

jasminerezai (approved):

PR #120, opened 2026-05-28
fix: generate schedule rows on activity creation + add week/month navigation

### LINKED ISSUE #119: fix: new activities not appearing in schedule + add week navigation
## Problem
When an admin/leader creates a new activity, it does not appear in the 
schedule. This is because `CREATE.newActivity()` only creates the 
`ActivityTemplate` and `TimeSlot` rows - it never inserts any `Schedule` 
rows. Schedule rows are currently only generated by the seeder.

Also flagged by supervisor: the schedule page has no navigation -
users cannot go to a different week or jump to a specific date.

## Changes
**Backend**
- `server/src/utils/weekCalculator.ts` - extract `nextDateForWeekday` 
  utility (currently only in seed.ts)
- `server/src/db/createQueries.ts` - generate Schedule rows for 
  current + next 2 weeks inside the same $transaction when a new 
  activity is created

**Frontend**
- `client/src/pages/SchedulePage.jsx` - new page replacing the 
  Placeholder on `/`, fetches real schedule data from the API
- `client/src/App.jsx` - wire up SchedulePage on the `/` route
- Week navigation: prev/next buttons + date picker to jump to any date

jasminerezai:
### Bug fix - new activities not appearing in schedule
When an admin/leader created a new activity, it would never appear in 
the schedule. The root cause was that `CREATE.newActivity()` only 
created the `ActivityTemplate` and `TimeSlot` rows but never inserted 
any `Schedule` rows. Schedule rows were previously only generated by 
the seeder.

Fix: `newActivity` now generates `Schedule` rows for the current + 
next 2 weeks inside the same `$transaction` when a new activity is 
created. If the schedule generation fails, the whole activity creation 
rolls back.

### Feature - week/month navigation on schedule page
The schedule page previously had no navigation - users were stuck on 
the current week/month with no way to look ahead or go back.

Added:
- ← Prev / Next → buttons for both weekly and monthly views
- Today button to jump back to current week/month
- Date picker to jump to any specific date (weekly view)
- Month label updates dynamically when navigating

## Files changed
- `server/src/utils/weekCalculator.ts` - extracted `nextDateForWeekday` utility
- `server/src/db/createQueries.ts` - generate Schedule rows on activity creation
- `client/src/pages/SchedulePage.jsx` - week/month navigation UI + fetch refetch on navigation

## Known limitation
Activities only appear in the schedule for the weeks that have been 
generated (~2-3 weeks ahead). Weeks further in the future will show 
empty until a cron job is implemented to generate schedule rows 
on a rolling weekly basis. That will be handled in a follow-up PR.

## Testing
- Create a new activity as admin/leader
- Navigate to the correct week - activity should appear immediately
- Prev/Next buttons navigate correctly
- Date picker jumps to the right week
- Monthly prev/next navigation works
- Today button returns to current week/month

## Closes
Closes #119

### DIFF
--- a/client/src/pages/SchedulePage.jsx
+++ b/client/src/pages/SchedulePage.jsx
@@ -41,6 +41,14 @@ const WEEKDAY_BY_INDEX = [
   'THURSDAY', 'FRIDAY', 'SATURDAY',
 ]
 
+function getMondayOfCurrentWeek() {
+  const d = new Date()
+  const day = d.getDay()
+  const diff = d.getDate() - day + (day === 0 ? -6 : 1)
+  d.setDate(diff)
+  d.setHours(0, 0, 0, 0)
+  return d
+}
 export default function SchedulePage() {
 
   // ── View State ────────────────────────────────────────────
@@ -150,14 +158,7 @@ export default function SchedulePage() {
 
   // ── Current Date ──────────────────────────────────────────
   const today = new Date()
-  const [currentWeekStart, setCurrentWeekStart] = useState(() => {
-    const d = new Date(today)
-    const day = d.getDay()
-    const diff = d.getDate() - day + (day === 0 ? -6 : 1)
-    d.setDate(diff)
-    d.setHours(0, 0, 0, 0)
-    return d
-  })
+  const [currentWeekStart, setCurrentWeekStart] = useState(getMondayOfCurrentWeek)
   const [currentMonth, setCurrentMonth] = useState(() => new Date())
 
   // ── Fetch Current Schedule ───────────────────────────────
@@ -491,11 +492,12 @@ export default function SchedulePage() {
 
   function handleDatePick(e) {
     if (!e.target.value) return
-    const picked = new Date(e.target.value)
-    const day = picked.getUTCDay()
-    const diff = day === 0 ? -6 : 1 - day
-    picked.setUTCDate(picked.getUTCDate() + diff)
-    picked.setUTCHours(0, 0, 0, 0)
+    const [y, m, d] = e.target.value.split('-').map(Number)
+    const picked = new Date(y, m - 1, d)
+    const day = picked.getDay()
+    const diff = picked.getDate() - day + (day === 0 ? -6 : 1)
+    picked.setDate(diff)
+    picked.setHours(0, 0, 0, 0)
     setCurrentWeekStart(picked)
   }
 
@@ -710,14 +712,7 @@ export default function SchedulePage() {
         {view === 'weekly' ? (
           <>
             <Button variant="outline" size="sm" onClick={prevWeek}>← Prev</Button>
-            <Button variant="ghost" size="sm" onClick={() => {
-              const d = new Date()
-              const day = d.getDay()
-              const diff = d.getDate() - day + (day === 0 ? -6 : 1)
-              d.setDate(diff)
-              d.setHours(0, 0, 0, 0)
-              setCurrentWeekStart(d)
-            }}>Today</Button>
+            <Button variant="ghost" size="sm" onClick={() => setCurrentWeekStart(getMondayOfCurrentWeek())}>Today</Button>
             <Button variant="outline" size="sm" onClick={nextWeek}>Next →</Button>
             <input
               type="date"
--- a/client/src/pages/SchedulePage.jsx
+++ b/client/src/pages/SchedulePage.jsx
@@ -49,6 +49,54 @@ function getMondayOfCurrentWeek() {
   d.setHours(0, 0, 0, 0)
   return d
 }
+
+function formatScheduleData(singleSchedule) {
+  const startDate = new Date(singleSchedule.startAt)
+  const formattedDate = `${startDate.getFullYear()}-${String(startDate.getMonth() + 1).padStart(2, '0')}-${String(startDate.getDate()).padStart(2, '0')}`
+
+  return {
+    id: singleSchedule.id,
+    activityId: singleSchedule.activityId,
+    title: singleSchedule.activity?.name || 'Activity',
+    sport: singleSchedule.activity?.name || 'Sport',
+    leader: singleSchedule.leaders
+      ?.map(leader => leader.profileName || 'Leader')
+      .join(', ') || 'Leader',
+    date: formattedDate,
+    time: startDate.toLocaleTimeString([], {
+      hour: '2-digit',
+      minute: '2-digit',
+    }),
+    location: singleSchedule.activity?.location || 'Unknown',
+    maxCapacity: singleSchedule.activity?.maxCapacity ?? null,
+    participantCount: singleSchedule.participantCount ?? 0,
+    cancelled: singleSchedule.status === 'CANCELLED',
+    notes: singleSchedule.activity?.notes || '',
+  }
+}
+
+function getWeeksForMonth(date) {
+  const year = date.getFullYear()
+  const month = date.getMonth()
+  const firstDay = new Date(year, month, 1)
+  const lastDay = new Date(year, month + 1, 0)
+
+  // Find the Monday of the week containing firstDay
+  const monday = new Date(firstDay)
+  const day = monday.getDay()
+  const diff = monday.getDate() - day + (day === 0 ? -6 : 1)
+  monday.setDate(diff)
+  monday.setHours(0, 0, 0, 0)
+
+  const weeks = []
+  const current = new Date(monday)
+  while (current <= lastDay) {
+    weeks.push(new Date(current))
+    current.setDate(current.getDate() + 7)
+  }
+  return weeks
+}
+
 export default function SchedulePage() {
 
   // ── View State ────────────────────────────────────────────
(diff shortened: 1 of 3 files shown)


server/src/db/createQueries.ts
```diff
@@ -44,45 +34,86 @@ export class CREATE {
...
                         }
                     }
                 }
+            });
+
+            // Generate Schedule rows for current + next 2 weeks
+            const today = new Date();
+            const scheduleData = [];
+
+            for (const slot of created.timeSlots) {
```
gunnarwrld, line 86: **`nextDateForWeekday` can produce a date in the _following_ week when `from` is already past the target weekday.**

`(target - current + 7) % 7` returns 0 when `from` is already on the target weekday, and returns a _forward_ offset otherwise. So if today is Wednesday (`current = 3`) and the slot is Monday (`target = 1`), `diff = (1 - 3 + 7) % 7 = 5`, placing the schedule row on the _next_ Monday, not the Monday of the current week.

This means activities created mid-week will have their early-week slots skipped for the current week and pushed to next week instead, producing incorrect schedule row dates.

Compute the Monday of `from`'s week first, then add the fixed weekday offset from that anchor:

```ts
export function nextDateForWeekday(from: Date, weekday: string): Date {
    const days = ['SUNDAY', 'MONDAY', 'TUESDAY', 'WEDNESDAY', 'THURSDAY', 'FRIDAY', 'SATURDAY'];
    const target = days.indexOf(weekday);
    const dow = from.getUTCDay();
    // Monday of from's ISO week
    const weekStart = new Date(from);
    weekStart.setUTCDate(from.getUTCDate() - ((dow + 6) % 7));
    // Add ISO weekday offset (Mon=0 ... Sun=6)
    const isoTarget = (target + 6) % 7;
    weekStart.setUTCDate(weekStart.getUTCDate() + isoTarget);
    return weekStart;
}
```

This will silently produce wrong data in production for any activity created on a day after the earliest slot weekday.

server/src/db/createQueries.ts
```diff
@@ -44,45 +34,86 @@ export class CREATE {
...
+                    },
+                    timeSlots: {
+                        createMany: {
+                            data: newAct.timeSlots.map(el => ({
+                                weekday: el.weekday,
+                                startTime: new Date(`1970-01-01T${el.startAt}Z`),
+                                endTime: new Date(`1970-01-01T${el.endAt}Z`)
+                            }))
+                        }
+                    },
```
gunnarwrld, line 61: **`week = 0` base date is today, not Monday of the current week, combined with the `nextDateForWeekday` bug this can produce schedule rows that duplicate or collide with later weeks.**

```ts
const base = new Date(today);
base.setUTCDate(today.getUTCDate() + week * 7);
```

`base` is today + N*7 days. When you then call `nextDateForWeekday(base, slot.weekday)`, for `week = 0` and `week = 1` with certain weekday/today combinations, the resolved date can land on the same calendar day (e.g., today is Friday, slot is Monday; week 0 resolves to _next_ Monday, week 1 also resolves to the Monday after that, skipping the current Monday entirely).

Add `base` to Monday of the current week, then add `week * 7`. Combined with the fix in `nextDateForWeekday`, this becomes:

```ts
const todayDow = today.getUTCDay();
const monday = new Date(today);
monday.setUTCDate(today.getUTCDate() - ((todayDow + 6) % 7));

for (const slot of created.timeSlots) {
    for (let week = 0; week < 3; week++) {
        const base = new Date(monday);
        base.setUTCDate(monday.getUTCDate() + week * 7);
        const date = nextDateForWeekday(base, slot.weekday);
        // ...
    }
}
```

server/src/db/createQueries.ts
```diff
@@ -44,45 +34,86 @@ export class CREATE {
...
+                    const date = nextDateForWeekday(base, slot.weekday);
+
+                    const start = new Date(date);
+                    start.setUTCHours(
+                        slot.startTime.getUTCHours(),
+                        slot.startTime.getUTCMinutes(),
+                        0, 0
+                    );
+
+                    const end = new Date(date);
```
gunnarwrld, line 100: **`tx.schedule.createMany` does not deduplicate. Re-running `newActivity` for an activity that somehow gets partially created (e.g., a retry after a timeout) will insert duplicate `Schedule` rows.**

Add a `skipDuplicates: true` flag:

```ts
await tx.schedule.createMany({ data: scheduleData, skipDuplicates: true });
```

This requires a unique constraint on `(activityId, startAt)` in `schema.prisma` if one doesn't already exist. Check `prisma/schema.prisma`, if it's missing, add:

```prisma
@@unique([activityId, startAt])
```

Without this, a partial failure + client retry will silently double-book schedule slots.

client/src/pages/SchedulePage.jsx
```diff
@@ -678,20 +705,46 @@ export default function SchedulePage() {
...
+                padding: '6px 10px',
+                border: '1.5px solid var(--color-border)',
+                borderRadius: 'var(--radius-sm)',
+                fontSize: 'var(--text-sm)',
+                fontFamily: 'var(--font-body)',
+                background: 'var(--color-surface-raised)',
+                color: 'var(--color-text)',
+                cursor: 'pointer',
+              }}
+            />
```
gunnarwrld, line 745: **"Today" button in the weekly view duplicates the Monday-of-current-week calculation that already exists in the `useState` initializer.**

The inline onClick:
```jsx
onClick={() => {
  const d = new Date()
  const day = d.getDay()
  const diff = d.getDate() - day + (day === 0 ? -6 : 1)
  d.setDate(diff)
  d.setHours(0, 0, 0, 0)
  setCurrentWeekStart(d)
}}
```

...is a copy-paste of the `useState` initializer at the top of the component. Extract a named helper `getMondayOfCurrentWeek()` and call it from both places. Otherwise the next person to touch the logic fixes one and misses the other.

client/src/pages/SchedulePage.jsx
```diff
@@ -162,8 +169,9 @@ export default function SchedulePage() {
         // Single fetch — API_BASE_URL is '' in dev so the
         // vite proxy handles it, and the deployed backend
         // origin in production.
+        const dateParam = `${currentWeekStart.getFullYear()}-${String(currentWeekStart.getMonth() + 1).padStart(2, '0')}-${String(currentWeekStart.getDate()).padStart(2, '0')}`
```
gunnarwrld: **`dateParam` is built with local-time methods, but the new endpoint `/api/schedules?date=...&entireWeek=true` presumably interprets it as a UTC week boundary on the server. Verify the server correctly handles local-date strings without timezone context.**

If the server's `startAndEndOfWeek` in `weekCalculator.ts` uses `setUTCDate`/`getUTCDate` (which it does), and the client sends a local-midnight Monday string, they will be in sync for UTC+ timezones but off-by-one for UTC- timezones (e.g., EST users navigating at 11pm will send Monday but the server treats it as Sunday UTC).

The safest fix is to append `T00:00:00Z` and always work in UTC on both sides, or send the ISO week number instead of a date string.

client/src/pages/SchedulePage.jsx
```diff
@@ -479,6 +472,40 @@ export default function SchedulePage() {
...
+
+  function nextWeek() {
+    setCurrentWeekStart(d => {
+      const n = new Date(d)
+      n.setDate(d.getDate() + 7)
+      return n
+    })
+  }
+
+  function handleDatePick(e) {
```
gunnarwrld, line 508: **`handleDatePick` uses UTC methods while the `useState` initializer and "Today" button use local-time methods. These will diverge by one day for users in timezones west of UTC.**

The initializer uses `d.getDay()` / `d.setDate()` (local time). `handleDatePick` uses `picked.getUTCDay()` / `picked.setUTCDate()` (UTC). Pick one convention. Since `dateParam` is built with local `getFullYear()`/`getMonth()`/`getDate()`, use local time throughout:

```js
function handleDatePick(e) {
  if (!e.target.value) return
  // ISO string 'YYYY-MM-DD' is parsed as UTC midnight -- force local via explicit constructor
  const [y, m, d] = e.target.value.split('-').map(Number)
  const picked = new Date(y, m - 1, d)
  const day = picked.getDay()
  const diff = picked.getDate() - day + (day === 0 ? -6 : 1)
  picked.setDate(diff)
  picked.setHours(0, 0, 0, 0)
  setCurrentWeekStart(picked)
}
```

gunnarwrld (changes requested):
Hi Jasmine! this PR has good structural decisions, the $transaction wrapping is correct and the navigation feature is well-scoped; but two correctness bugs in the schedule-row generation will silently produce wrong data in production and must be fixed before merge.

server/src/db/createQueries.ts
JuriRappold: > `(target - current + 7) % 7` returns 0 when `from` is already on the target weekday

@gunnarwrld towards the part of your comment concerning when `from` and `weekday` are the same weekday, the function description seems to use this as the intended behavior:
> If the base date IS that weekday, returns the same day.
---
But I agree on the forward offset issue, the function description says:
> returns the date of that weekday in the same week as the base date.

In that sense, it is a bug in the function for it to only have a forward offset. Especially in its current use case. Also I think the name is slightly confusing and probably part of the reason for this Bug. A more apropiate name could be something like `getDateOfWeekday(from: Date, weekday: string)` with a description like this:
```
/**
* Returns date of the weekday in the week from the given Date.
* @param from any date functioning as a base
* @param weekday ['SUNDAY', 'MONDAY', 'TUESDAY', 'WEDNESDAY', 'THURSDAY', 'FRIDAY', 'SATURDAY']
* @return date of the weekday from the week of the base date (from-parameter)
*/
```

---
Btw, nice calculation of the start of the week: `weekStart.setUTCDate(from.getUTCDate() - ((dow + 6) % 7));`. Probably a more reliable way to calculate the start and end of the week than the function I wrote early on. Might want to replace it:
```typescript
export function startAndEndOfWeek(nowDate: Date) {
    return {
        startDay: new Date( nowDate.getUTCDate() - ( (nowDate.getUTCDay() + 6) % 7 ) ),
        endDay: new Date( nowDate.getUTCDate() + ( (nowDate.getUTCDay() + 6) % 7 ) )
    }
}
```

server/src/utils/weekCalculator.ts
```diff
@@ -38,4 +38,20 @@ export function lastWeek(date: Date, numberOfWeeks: number): Date {
...
     return result;
+}
+
+
+/**
+ * Given a base date and a Weekday enum string (e.g. 'MONDAY'),
+ * returns the date of that weekday in the same week as the base date.
+ * If the base date IS that weekday, returns the same day.
+ */
+export function nextDateForWeekday(from: Date, weekday: string): Date {
```
JuriRappold, line 48: See discussion in the `createQueries.ts` for the bugs

client/src/pages/SchedulePage.jsx
JuriRappold: The Backend accepts `YYYY-MM-dd`, *iso strings*, or the `"today"`, but without a parameter it defaults to the current date with ISO string:
```typescript
export const ScheduleDateSchema = z.nullish(z.union([
    z.string().date(),
    z.string().datetime(),
    z.literal("today")
])).default(() => new Date().toISOString());
```
--> See *schedule.controller.ts* `getSchedule()` function
[Zod Date Docs](https://zod.dev/api#iso-dates)

JuriRappold (changes requested):
Good Job! Only the `nextDateForWeekday` function doesn't work for its intended and current use case. Gunnar already flagged that.

JuriRappold (approved):
Looks good to me!

client/src/pages/SchedulePage.jsx
```diff
@@ -150,8 +158,8 @@ export default function SchedulePage() {
 
   // ── Current Date ──────────────────────────────────────────
   const today = new Date()
-  // Test for JUNE
-  //const today = new Date('2026-06-15')
+  const [currentWeekStart, setCurrentWeekStart] = useState(getMondayOfCurrentWeek)
+  const [currentMonth, setCurrentMonth] = useState(() => new Date())
 
   // ── Fetch Current Schedule ───────────────────────────────
   useEffect(() => {
```
gunnarwrld, line 213: The  useEffect  re-fetches when  `currentWeekStart`  changes, but the fetch URL always uses  `dateParam`  derived from  `currentWeekStart` . When the user is in monthly view and clicks Prev/Next month,  `currentMonth`  changes but  `currentWeekStart`  does not, so  `fetchSchedule`  never fires and the monthly view continues showing stale data from whatever week was last fetched.

This is a logic gap. Either:
•	Add  currentMonth  to the dependency array and conditionally build the fetch URL based on  view , or
•	Keep two separate  useEffect  hooks, one per view

The current URL already uses  `entireWeek=true`  which implies it is only designed for weekly fetching. If the monthly view is supposed to show data from the same weekly API response then the approach needs rethinking, the monthly view for a full month needs a different API call or client-side aggregation across multiple weekly fetches.

gunnarwrld (changes requested):
One more thing to goooo!!

larrymijo (approved):
Great job!

client/src/pages/SchedulePage.jsx
```diff
@@ -41,6 +41,62 @@ const WEEKDAY_BY_INDEX = [
...
+    }),
+    location: singleSchedule.activity?.location || 'Unknown',
+    maxCapacity: singleSchedule.activity?.maxCapacity ?? null,
+    participantCount: singleSchedule.participantCount ?? 0,
+    cancelled: singleSchedule.status === 'CANCELLED',
+    notes: singleSchedule.activity?.notes || '',
+  }
+}
+
+function getWeeksForMonth(date) {
```
gunnarwrld, line 78: `getWeeksForMonth` returns the Monday of every ISO week that overlaps with the month. For a month like October 2026 where the 1st is Thursday, the first Monday anchor is September 28 (that week's fetch includes Sept 28–Oct 4), meaning you are fetching data from the previous month. More practically: a month spanning 6 partial weeks fires 6 concurrent fetch calls every time `currentMonth` changes. For the current small dataset this is fine, but it will also re-fetch the same data every time view changes while in monthly mode (since view is now a dependency).

Consider maybe doing something like this:
`}, [view, currentWeekStart, currentMonth])`
When view flips from 'weekly' to 'monthly', this fires the monthly `Promise.all` even if `currentMonth` hasn't changed. Cache the last fetched month in a ref, or split into two `useEffect` hooks keyed on their own state:
```
jsx
useEffect(() => { /* weekly fetch */ }, [currentWeekStart])
useEffect(() => { /* monthly fetch */ }, [currentMonth])
// view switch just re-renders, no new fetch needed if data is already loaded
```

client/src/pages/SchedulePage.jsx
```diff
@@ -150,120 +206,86 @@ export default function SchedulePage() {
...
-  // Test for JUNE
-  //const today = new Date('2026-06-15')
+  const [currentWeekStart, setCurrentWeekStart] = useState(getMondayOfCurrentWeek)
+  const [currentMonth, setCurrentMonth] = useState(() => new Date())
 
   // ── Fetch Current Schedule ───────────────────────────────
   useEffect(() => {
+    let active = true
 
     async function fetchSchedule() {
```
gunnarwrld, line 216: Each individual week fetch inside `fetchPromises` does not check active. If the component unmounts mid-flight, `Promise.all` still resolves and `setActivities` is called. Move the active check inside each promise, or abort via `AbortController`:

For now the `let active = true / return () => { active = false }` pattern is already present from the weekly branch, just apply it consistently inside the individual week fetches too, or use a single `AbortController` for the whole effect.

gunnarwrld (approved):
The backend changes (`createQueries.ts`, `weekCalculator.ts`) are correct and clean. The core feature works. Issues #1 and #2 above are quality/performance concerns rather than correctness blockers given the current dataset size, if you guys are is comfortable carrying them as follow-up tech debt, this is mergeable!

PR #122, opened 2026-05-28
feat: generate 16 weeks of schedules on startup, cron generates 2 weeks ahead weekly

### LINKED ISSUE #121: feat: cron job to generate schedule rows weekly
## Problem
Schedule rows are currently only generated when a new activity is 
created (~3 weeks ahead). Any week further out shows empty even 
if activities exist. There is no automated process to keep the 
schedule rolling forward.

## Solution
Implement a cron job that runs every Monday and generates Schedule 
rows for the upcoming week for all active ActivityTemplates that 
have TimeSlots.

On server startup, generate 16 weeks ahead (~4 months) so the 
schedule is always populated and navigable from day one.

## Requirements
- On startup: generate Schedule rows for the next 16 weeks for 
  all active ActivityTemplates with TimeSlots
- Weekly cron: runs every Monday at 00:05, generates 1 week ahead
- Skip any rows that already exist (skipDuplicates: true)
- Use nextDateForWeekday utility from weekCalculator.ts
- Use defaultStatus from the ActivityTemplate for each row

## Why 16 weeks
The schedule page supports navigation months ahead. 16 weeks (~4 
months) ensures the teacher and users always see a populated 
schedule when navigating forward without waiting for the cron.

jasminerezai:
## What this PR does

Extends the existing cron job to fully populate the schedule on 
startup and keep it rolling forward automatically.

### Changes

**On server startup:**
Generates Schedule rows for the next 16 weeks (~4 months ahead) 
for all active ActivityTemplates. This ensures the schedule is 
always fully populated from day one - Deema and users can 
navigate months ahead without seeing empty weeks.

**Weekly cron (every Monday 00:05):**
Now generates 2 weeks ahead instead of 1, so there's always a 
buffer even between startups.

**Both are safe to run multiple times** - existing rows are 
skipped automatically.

## Files changed
- `server/src/cron/scheduleGenerator.ts` - added `weeksAhead` 
  parameter, loops over multiple weeks in a single run
- `server/src/cron/index.ts` - startup now calls 
  `generateWeeklySchedules(16)`, cron calls 
  `generateWeeklySchedules(2)`

## Testing
Restart the backend and watch the terminal - you should see:
[cron] Generating schedules for the next 16 weeks...
[cron] Week 1/16 - X created, X skipped.
...
[cron] Total - 298 schedules created, 6 already existed.

## Note
This PR depends on #119 for the frontend navigation to be visible. 
So I'll merge #119 first before merging this one.

## Closes
Closes #121


JuriRappold (approved):
Good job! Don't see any issues

server/src/cron/scheduleGenerator.ts
```diff
@@ -32,86 +26,75 @@ function nextDateForWeekday(base: Date, weekday: string): Date {
...
-                },
-            });
-
-            if (existing) {
-                skipped++;
-                continue;
+    for (let week = 1; week <= weeksAhead; week++) {
+        const futureDate = nextWeek(now, week);
+        const { startDay, endDay } = startAndEndOfWeek(futureDate);
+
```
gunnarwrld, line 70: **N+1 Prisma queries inside the week loop -- this will fire `activities * slots * weeksAhead` sequential DB round-trips on startup (16 weeks * N slots = potentially hundreds of individual queries).**

The existing `findFirst` per slot was flagged as a known TODO in the original code. This PR multiplies that problem by 16. For startup with 16 weeks and even a modest 20 activities × 3 slots, that is 960 sequential `findFirst` calls before a single `create`.

Fetch all existing schedules for the entire date range in one query upfront, build a Set for O(1) lookup, then batch-insert with `createMany`:

```ts
export async function generateWeeklySchedules(weeksAhead: number = 1): Promise<void> {
    const now = new Date();
    const activities = await prisma.activityTemplate.findMany({
        where: { defaultStatus: ActivityStatus.ACTIVE },
        include: { timeSlots: true },
    });

    // Compute the full date range we're about to generate
    const rangeStart = startAndEndOfWeek(nextWeek(now, 1)).startDay;
    const rangeEnd = startAndEndOfWeek(nextWeek(now, weeksAhead)).endDay;

    // Fetch all existing schedule rows in range once
    const existing = await prisma.schedule.findMany({
        where: { startAt: { gte: rangeStart, lte: rangeEnd } },
        select: { activityId: true, startAt: true },
    });
    const existingSet = new Set(
        existing.map(s => `${s.activityId}|${s.startAt.toISOString()}`)
    );

    const toCreate: Prisma.ScheduleCreateManyInput[] = [];

    for (let week = 1; week <= weeksAhead; week++) {
        const { startDay } = startAndEndOfWeek(nextWeek(now, week));
        for (const activity of activities) {
            for (const slot of activity.timeSlots) {
                const slotDate = nextDateForWeekday(startDay, slot.weekday);
                const startAt = new Date(slotDate);
                startAt.setUTCHours(slot.startTime.getUTCHours(), slot.startTime.getUTCMinutes(), 0, 0);
                const endAt = new Date(slotDate);
                endAt.setUTCHours(slot.endTime.getUTCHours(), slot.endTime.getUTCMinutes(), 0, 0);

                if (existingSet.has(`${activity.id}|${startAt.toISOString()}`)) continue;
                toCreate.push({ activityId: activity.id, startAt, endAt, status: ActivityStatus.ACTIVE });
            }
        }
    }

    await prisma.schedule.createMany({ data: toCreate, skipDuplicates: true });
    console.log(`[cron] Total — ${toCreate.length} schedules created.`);
}
```

This reduces N*M*weeksAhead queries down to 2 queries total regardless of dataset size.

server/src/cron/scheduleGenerator.ts
```diff
@@ -32,86 +26,75 @@ function nextDateForWeekday(base: Date, weekday: string): Date {
...
+        console.log(`[cron] Generating schedules for ${startDay.toDateString()} – ${endDay.toDateString()}`);
+
+        let created = 0;
+        let skipped = 0;
+
+        for (const activity of activities) {
+            for (const slot of activity.timeSlots) {
+
+                const slotDate = nextDateForWeekday(startDay, slot.weekday);
```
gunnarwrld, line 81: **`prisma.schedule.create` (singular) inside the inner loop compounds the N+1 problem and bypasses duplicate safety.**

Even if you keep the `findFirst` guard, issuing individual `create` calls inside a loop means each insert is a separate round-trip with no transaction boundary. If the process is killed mid-run on startup, you'll have a partially populated schedule with no way to tell which weeks were completed.

At minimum, collect rows into an array and call `createMany` once per week, or once total (see the refactor above). `createMany` with `skipDuplicates: true` makes the `findFirst` check redundant and eliminates the race window between check and insert.

server/src/cron/scheduleGenerator.ts
```diff
@@ -19,10 +17,6 @@ const weekdayMap: Record<string, number> = {
     SUNDAY: 0,
 };
 
-// nextDateForWeekday
-// Given a base date and a weekday, returns the next occurrence
-// of that weekday on or after the base date.
-// Reuses the same logic as seed.ts.
 function nextDateForWeekday(base: Date, weekday: string): Date {
     const target = weekdayMap[weekday]!;
     const date = new Date(base);
```
gunnarwrld, line 26: `(target - current + 7) % 7` returns the _next_ occurrence of the weekday from `base`, not the occurrence within `base`'s ISO week. Since `base` is `startDay` (Monday of the target week), this is actually safe here; `startDay` is always Monday (day 1), so `diff` is always `(target - 1 + 7) % 7` which correctly gives 0 for Monday, 1 for Tuesday, etc.

However, this safety is implicit and depends on `startDay` always being a Monday. If `startAndEndOfWeek` ever returns a different anchor (e.g. due to a DST edge case or refactor), the bug silently reappears. 

Add an assertion or a comment making the Monday-anchor assumption explicit:

```ts
// startDay is always Monday -- nextDateForWeekday is safe here
// because (target - 1 + 7) % 7 gives the correct in-week offset.
const slotDate = nextDateForWeekday(startDay, slot.weekday);
```

Otherwise we could fix: `nextDateForWeekday` itself in `weekCalculator.ts`  so callers don't need to reason about this invariant.

gunnarwrld (changes requested):
This PR is conceptually correct, extending `generateWeeklySchedules` to accept `weeksAhead` is the right abstraction, and the startup pre-population strategy is great. However, the implementation multiplies a pre-existing N+1 query problem by 16x, which is the blocking issue. Everything else can be addressed in a follow-up or alongside the fix.

JuriRappold (approved):
Good catch Gunnar! And Nicely solved Jasmine!

server/src/cron/scheduleGenerator.ts
```diff
@@ -32,86 +29,77 @@ function nextDateForWeekday(base: Date, weekday: string): Date {
...
     const activities = await prisma.activityTemplate.findMany({
         where: { defaultStatus: ActivityStatus.ACTIVE },
         include: { timeSlots: true },
     });
 
-    let created = 0;
-    let skipped = 0;
+    // 2. Compute the full date range we're about to generate
+    const rangeStart = startAndEndOfWeek(nextWeek(now, 1)).startDay;
+    const rangeEnd = startAndEndOfWeek(nextWeek(now, weeksAhead)).endDay;
```
gunnarwrld: `rangeEnd` calculation is off-by-one for multi-week ranges. `startAndEndOfWeek(nextWeek(now, weeksAhead))` gives the end of week N. But when `weeksAhead = 16`, `nextWeek(now, 16)` returns the Monday 16 weeks from now, and `endDay` is the Sunday of that same week. That's correct, but only if `nextWeek(now, n)` is defined as "the Monday of the nth week ahead." Verify this against `weekCalculator.ts`. If `nextWeek(now, 1)` returns "next Monday" and `nextWeek(now, 2)` returns "the Monday two weeks out," the range is correct. Otherwise it always returns the same week regardless of n, `rangeEnd` will be wrong and the existing-schedule pre-fetch will silently under-filter, causing false duplicates to be inserted (caught by `skipDuplicates: true` as a safety net, but it means the Set check is broken).

Maybe add a test or a quick log to validate: `rangeEnd` should be exactly `7 * weeksAhead` days after `rangeStart`.

server/src/cron/scheduleGenerator.ts
```diff
@@ -32,86 +29,77 @@ function nextDateForWeekday(base: Date, weekday: string): Date {
...
-
-            created++;
+                });
+            }
         }
     }
 
-    console.log(`[cron] Done — ${created} schedules created, ${skipped} already existed.`);
+    // 5. Batch insert all missing rows in one query
+    await prisma.schedule.createMany({ data: toCreate, skipDuplicates: true });
```
gunnarwrld: skipDuplicates: true is load-bearing, not just a safety net. The Set dedup is based on `activityId | startAt.toISOString(). toISOString()` is millisecond-precise. The existing rows are fetched from the DB where `startAt` is stored as a `TIMESTAMPTZ`. Depending on how Prisma deserializes the stored value (e.g., microseconds truncated vs. not), there can be a mismatch between the stored value's ISO string and the newly computed `startAt.toISOString()`. This would cause the Set check to miss a match and attempt an insert, which `skipDuplicates: true` would then silently absorb.

This is not a crash risk  (cause `skipDuplicates` its our superman lol) but the logged count (`toCreate.length`) would be wrong (it counts rows you tried to insert that already existed). To make the Set reliable, normalize both sides, and ensure `startAt` is always constructed via setUTCHours with integer values only (which this PR already does correctly). The risk is on the DB read side, not the write side.

gunnarwrld (changes requested):
One real issues to fix before merging: validating that the `rangeEnd` calculation is correct for `nextWeek(now, n)` semantics. The timestamp normalization in the Set is low-risk but worth addressing for correctness of the logged count. Everything else is minor. The algorithmic core is correct and a genuine improvement over the prior implementation, Good job!!

server/src/cron/scheduleGenerator.ts
```diff
@@ -31,87 +28,93 @@ function nextDateForWeekday(base: Date, weekday: string): Date {
...
+
+    console.log(`[cron] Done — ${toCreate.length} schedules created, ${existing.length} already existed.`);
 }
```
gunnarwrld: `skipDuplicates: true` on `createMany` is a silent safety net but it hides real duplicates from your log. If `existingSet` deduplication logic has a bug (e.g. the key normalization is wrong), rows will silently not be inserted and `toCreate.length` will misreport how many were actually written. The final log line says `${toCreate.length} schedules created` but `createMany` returns `{ count: N }`, use that returned count instead so the log reflects what Prisma actually inserted, not what you sent.

```ts
const result = await prisma.schedule.createMany({ data: toCreate, skipDuplicates: true });
console.log(`[cron] Done -- ${result.count} schedules created, ${existing.length} already existed.`);
```

server/src/cron/scheduleGenerator.ts
```diff
@@ -31,87 +28,93 @@ function nextDateForWeekday(base: Date, weekday: string): Date {
...
 
+// normalizes a date for reliable deduplication, zeroing seconds and milliseconds
+function getDeduplicationKey(activityId: string, date: Date): string {
+    const normalized = new Date(date);
+    normalized.setUTCSeconds(0, 0);
+    return `${activityId}|${normalized.getTime()}`;
+}
+
 // generateWeeklySchedules
```
gunnarwrld, line 37: `getDeduplicationKey` zeroes seconds and milliseconds but NOT minutes. A DB row with `startAt = 09:01:00.000Z` and a generated slot with `startAt = 09:00:00.000Z` produce different keys, so the duplicate is not caught. Given that `startAt` is always set with `setUTCHours(h, m, 0, 0)` this is safe today, but the comment says it normalizes for 'millisecond/microsecond variations'. The normalization should be consistent with what the generator actually stores, or better: key on `${activityId}|${date.getUTCFullYear()}-${date.getUTCMonth()}-${date.getUTCDate()}-${date.getUTCHours()}-${date.getUTCMinutes()}` to make the intent explicit and resilient to future changes to precision.

server/src/__tests__/scheduleGenerator.test.ts
```diff
@@ -0,0 +1,156 @@
...
+  it('tolerates DB millisecond/microsecond variations and skips duplicate generation', async () => {
+    const activityId = testUUID();
+    const now = new Date();
+
+    // Let's find what the generated Monday date would be
+    // We can simulate nextMonday start date
+    // We'll mock activity templates with a TUESDAY time slot
+    mockPrismaClient.activityTemplate.findMany.mockResolvedValueOnce([
+      {
+        id: activityId,
```
gunnarwrld, line 157: The third test re-imports `weekCalculator.js` at runtime inside the test body with `await import(...)`. This works but couples the test's expected value computation to the same utility under test. If `nextWeek` or `startAndEndOfWeek` are broken, this test will silently pass because both the SUT and the expected value use the same broken function. Compute the expected `baseMonday` manually (raw date arithmetic) to make the test independent:

```ts
// Get next Monday from now without using the utility under test
const now = new Date();
const daysUntilMonday = (8 - now.getUTCDay()) % 7 || 7;
const baseMonday = new Date(Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate() + daysUntilMonday));
```

server/src/__tests__/scheduleGenerator.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    expect(diffDays).toBe(7);
+
+    // Verify createMany was called with 1 schedule row
+    expect(mockPrismaClient.schedule.createMany).toHaveBeenCalledTimes(1);
+    const createArgs = vi.mocked(mockPrismaClient.schedule.createMany).mock.calls[0]?.[0];
+    expect(createArgs).toBeDefined();
+    expect(createArgs!.data).toHaveLength(1);
+    expect((createArgs!.data as any)[0].activityId).toBe(activityId);
+  });
+
```
gunnarwrld, line 77: The 16-week test asserts `createArgs!.data` has length 16, but the mock has one activity with one Monday slot repeated over 16 weeks (that is correct). However, there is no assertion on `rangeStart` or `rangeEnd` values, only on the span. A off-by-one in `weeksAhead` calculation (e.g. starting from week 0 instead of week 1) would still pass this test. Add at least one assertion that `rangeStart` is the Monday of next week, not today or two weeks from now.

server/src/cron/scheduleGenerator.ts
```diff
@@ -31,87 +28,93 @@ function nextDateForWeekday(base: Date, weekday: string): Date {
...
+    const rangeStart = startAndEndOfWeek(nextWeek(now, 1)).startDay;
+    // Calculate rangeEnd directly based on weeksAhead * 7 days to avoid any DST, timezone, or day-of-week calculation errors.
+    // Sunday end is Monday 00:00:00 + (weeksAhead * 7 days) - 1 ms.
+    const rangeEnd = new Date(rangeStart.getTime() + weeksAhead * 7 * 24 * 60 * 60 * 1000 - 1);
 
-    for (const activity of activities) {
-        for (const slot of activity.timeSlots) {
+    // Validate the date range spans exactly weeksAhead * 7 days
+    const diffDays = Math.round((rangeEnd.getTime() - rangeStart.getTime() + 1) / (24 * 60 * 60 * 1000));
+    if (diffDays !== weeksAhead * 7) {
```
gunnarwrld: Runtime date range validation with `console.error` is not a substitute for a guard. If `diffDays !== weeksAhead * 7` you log a warning and continue anyway, potentially inserting schedules into a malformed range. This edge case should never happen given the arithmetic, but if it does the correct action is to throw and abort, not log and proceed. Either remove the check entirely (trust the math) or make it a hard throw:

```ts
if (diffDays !== weeksAhead * 7) {
    throw new Error(`[cron] Date range mismatch: expected ${weeksAhead * 7} days, got ${diffDays}`);
}
```

gunnarwrld (changes requested):
The refactor is architecturally correct. N+1 Prisma queries are gone, the in-memory Set deduplication is the right pattern, and `createMany` batch insert is the correct tool here. Some few issues need to be addressed before merging.

server/src/cron/scheduleGenerator.ts
```diff
@@ -31,87 +28,91 @@ function nextDateForWeekday(base: Date, weekday: string): Date {
     return date;
 }
 
+// normalizes a date for reliable deduplication, keying on explicit date parts to be resilient to precision variations
```
gunnarwrld, line 32: **`getDeduplicationKey` truncates precision at minutes but does not zero-pad month, day, hour, or minute.** `getUTCMonth()` returns 0–11, `getUTCDate()` returns 1–31, `getUTCHours()` 0–23, `getUTCMinutes()` 0–59 -- all without padding. This means `9:05` and `9:50` both produce the same minute segment (`9-5` vs `9-50` -- fine), but more dangerously `month 9 day 1` and `month 9 day 10` will collide if hours/minutes happen to share a prefix pattern in the string.

Use ISO-style zero-padded fields, or just use `toISOString().slice(0, 16)`:
```ts
function getDeduplicationKey(activityId: string, date: Date): string {
    return `${activityId}|${date.toISOString().slice(0, 16)}`;
}
```
This is less error-prone and immediately readable.

server/src/cron/scheduleGenerator.ts
```diff
@@ -31,87 +28,91 @@ function nextDateForWeekday(base: Date, weekday: string): Date {
...
-
-            created++;
+                });
+            }
         }
     }
 
-    console.log(`[cron] Done — ${created} schedules created, ${skipped} already existed.`);
+    // 5. Batch insert all missing rows in one query
+    const result = await prisma.schedule.createMany({ data: toCreate, skipDuplicates: true });
```
gunnarwrld, line 110: **`skipDuplicates: true` on `createMany` is redundant given the `existingSet` check above.** Both guard against duplicates, the in-memory set already filters them before this call. `skipDuplicates` relies on a DB-level unique constraint existing on `(activityId, startAt)`. If that constraint exists, you do not need the in-memory set; if it does not exist, `skipDuplicates` silently swallows inserts without error. Pick one:
- Keep the in-memory `existingSet` dedup (current approach) and remove `skipDuplicates: true`, which makes the intent explicit and does not depend on a DB constraint.
- Or drop the in-memory set and rely solely on `skipDuplicates: true`, which is simpler but requires the unique constraint to be present in the schema.

Also verify `schema.prisma` has `@@unique([activityId, startAt])` on `Schedule`; if it does not, `skipDuplicates` is silently a no-op.

server/src/cron/index.ts
```diff
@@ -1,37 +1,27 @@
...
+    // Startup: generate 16 weeks ahead (~4 months) so the schedule
+    // is always fully populated from day one, regardless of what day
+    // the server starts on.
+    console.log('[cron] Generating schedules for the next 16 weeks...');
+    generateWeeklySchedules(16).catch(err => {
+        console.error('[cron] Startup schedule generation failed:', err);
+    });
 
     console.log('[cron] Weekly schedule generation cron job registered (Monday 00:05).');
 }
```
gunnarwrld: Both `cron/index.ts` and `cron/scheduleGenerator.ts` use `console.log` and `console.error` extensively. These are operational logs so they are more acceptable than debug logs, but for this project we enforced no `console.log` in production paths. Consider using a proper logger or at minimum a named prefix wrapper. At a minimum the startup `console.log('[cron] Generating schedules...')` on line 27 fires synchronously before the async call completes, so it does not accurately reflect progress.

gunnarwrld: **Missing newline at EOF** in both `cron/index.ts` and `cron/scheduleGenerator.ts`. Add trailing newlines.

server/src/__tests__/scheduleGenerator.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    expect(rangeStart).toBeDefined();
+    expect(rangeEnd).toBeDefined();
+
+    // Verify rangeStart is the Monday of next week
+    const now = new Date();
+    const daysUntilMonday = (8 - now.getUTCDay()) % 7 || 7;
+    const expectedRangeStart = new Date(Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate() + daysUntilMonday));
+    expect(rangeStart.getTime()).toBe(expectedRangeStart.getTime());
+
```
gunnarwrld, line 53: **Test computes `daysUntilMonday` independently from the code under test, this is correct practice.** The deduplication key collision test (test 3) is the most valuable one here; the range-boundary tests are good as well. One gap: none of the tests verify the actual `startAt`/`endAt` timestamps of the generated rows (only that `data` has the right length and `activityId`). A future test should assert that the generated `startAt` matches the expected UTC datetime for the slot, to guard against time zone bugs in `setUTCHours`.

server/src/cron/scheduleGenerator.ts
```diff
@@ -31,87 +28,91 @@ function nextDateForWeekday(base: Date, weekday: string): Date {
...
+    const rangeStart = startAndEndOfWeek(nextWeek(now, 1)).startDay;
+    // Calculate rangeEnd directly based on weeksAhead * 7 days to avoid any DST, timezone, or day-of-week calculation errors.
+    // Sunday end is Monday 00:00:00 + (weeksAhead * 7 days) - 1 ms.
+    const rangeEnd = new Date(rangeStart.getTime() + weeksAhead * 7 * 24 * 60 * 60 * 1000 - 1);
 
-    for (const activity of activities) {
-        for (const slot of activity.timeSlots) {
+    // Validate the date range spans exactly weeksAhead * 7 days
+    const diffDays = Math.round((rangeEnd.getTime() - rangeStart.getTime() + 1) / (24 * 60 * 60 * 1000));
+    if (diffDays !== weeksAhead * 7) {
```
gunnarwrld: **Runtime date-range assertion in production code.** The `diffDays !== weeksAhead * 7` guard throws an `Error` that will bubble up and crash the startup cron if it ever triggers. The invariant here is a pure arithmetic consequence of how `rangeEnd` is constructed (`rangeStart + weeksAhead * 7 * 86400000 - 1`), it can only fail if `rangeStart` itself is `Invalid Date`, which would already be caught earlier.

This check is more appropriate as a unit test assertion than a runtime throw. Remove it from production code, or at minimum catch and log rather than throw, so the server does not crash on startup over a scheduling edge case.

gunnarwrld (changes requested):
This is a solid architectural improvement, eliminating N*M sequential findFirst calls per slot in favour of a single findMany + in-memory Set lookup is the right call, and the weeksAhead parameter makes the generator reusable cleanly. The test coverage for deduplication and range boundaries is appreciated. Just some more things to leave the code in excelent form.

gunnarwrld (approved):
One remaining note worth flagging: the `startAt` timestamp assertions gap in the tests (no test verifies the actual UTC datetime of generated rows) is still technically open as a future improvement, but it is not a blocker.

With that said, this PR is ready to merge. No open threads, the core N+1 architectural concern is fixed, and the implementation is correct.

